PR #2, opened 2020-01-16
Add data and converter script for Tomczak et. al. (2013) GSMF

calvin-sykes:
This dataset tabulates the GSMF at multiple redshifts. The script will produce an HDF5 file for each of these, named like Tomczak2013_z001p250.hdf5.


data/GalaxyStellarMassFunction/conversion/convertTomczak2013.py
```diff
@@ -0,0 +1,157 @@
...
+    z: the redshift to produce the GSMF for. The given value corresponds to the lower edge of a range in redshift, which has width 0.25 for z < 1.5 and 0.5 for z >= 1.5
+    mstar_bins: the list of stellar mass bins for which the GSMF is tabulated
+    gsmf_at_z: the slice of the GSMF array at the chosen redshift
+    """
+
+    processed = ObservationalData()
+
+    comment = f"Assuming Chabrier IMF, quoted redshift is lower bound of range. h-corrected for SWIFT using Cosmology: {cosmology.name}."
+    citation = "Tomczak et al (2013)"
+    bibcode = " 2014ApJ...783...85T"
```
JBorrow: Please remove the leading space here

data/GalaxyStellarMassFunction/conversion/convertTomczak2013.py
```diff
@@ -0,0 +1,157 @@
...
+    name = "GSMF from ZFOURGE/CANDELS"
+    plot_as = "points"
+    redshift = z
+    h = cosmology.h
+
+    M = 10 ** mstar_bins * unyt.Solar_Mass
```
JBorrow: I think we want to un-h-correct the data and then re-correct it - as far as I am aware this should be corrected as follows:

M -> M * (h_new / h_old)**-2
Phi -> Phi * (h_new / h_old)**3

data/GalaxyStellarMassFunction/conversion/convertTomczak2013.py
```diff
@@ -0,0 +1,157 @@
...
+import unyt
+import numpy as np
+import os
+import re
+import sys
+
+
+def handle_bad_values(parts):
+    """
+    Convert "missing data" values to NaNs and fix the "errors" on these values getting misinterpreted as additional data points.
```
JBorrow: Can you make sure that the docstrings have a max length of 88 please?

JBorrow:
Final minor comment - don't forget the formatting script `format_all.sh` at the top level.

calvin-sykes:
Should be all sorted now.

data/GalaxyStellarMassFunction/conversion/convertTomczak2013.py
calvin-sykes: Just a thought: if anyone ever wanted to have the same data for different cosmologies, they couldn't at the moment (without manual intervention). If that's a likely use case, would it be better to adapt to the desired cosmology when retrieving the data to plot it?

JBorrow: Do you mean the manual intervention of changing the cosmology.py file and re-running? That is perhaps a concern, but the number of cosmologies we use is ~3 so I was just planning on creating three copies of the reduced data.

calvin-sykes: I meant manually moving/renaming the data files so that they don't get overwritten when you rerun with a different cosmology.
I guess an easy solution would be to include the cosmology name in the filename for the hdf5 file, or having a separate directory structure per cosmology?

JBorrow: My 'easy solution' to this is to clone the repository n times, changing the top level name (e.g. comparison-data-planck15) and then change the cosmology.py each time... However I will admit I haven't put much thought into this specific issue. I worry that adding more layers of folders will just make things confusing (they're already quite deep).

Do you think the above easy solution is a bad idea?

calvin-sykes: No, I guess that works (It's the same amount of folders though, just arranged differently!)

PR #4, opened 2020-01-17
Added Trujillo2020 sizes sand Crain2015

JBorrow:
@MatthieuSchaller Can you check that my un-correction and re-correction of the EAGLE size data is correct?


MatthieuSchaller:
Where did you get these files from? I can only tell you how to correct if I know how they have been obtained.

JBorrow:
They're the ones that you sent me

MatthieuSchaller:
Then no correction at all.

MatthieuSchaller:
The data you have here from EAGLE has been corrected to be h-free. It is identical to the result you would have obtained in a simulation code in which there never was any h in the units or anywhere in the code to start with.

 - When observational data is given in Mpc/h then we need to convert to Mpc.
 - When observational data is given in Mpc, no need to correct.
 - When observational data is given in Mpc/h70 then we need to convert to Mpc.

In that latter case, the correction is not 0.7 (say) but the ratio of their h to the one you want in the cosmology we use.

PR #9, opened 2020-01-20
Add Deshmukh et. al. 2018 GSMF data and conversion script

calvin-sykes:
Last one for now I promise :laughing:

This is for z=2-6, so probably mostly useful for me at the moment. I figured I'd submit it anyway.


JBorrow:
As this is also a high redshift gsmf I'll give you the heads up that I'm trying to get other high-redshift simulation comparison data.

PR #12, opened 2020-01-24
Add data and conversion scripts for all the SFR functions used in Kat…

calvin-sykes:
Added a bunch of star formation rate functions. The data is taken from the tables in Katsianis et al (2016, 2017) but I've used the original papers for the citations.


JBorrow:
Thanks Calvin!

PR #17, opened 2020-03-21
Add various metallicity - stellar mass relation

MatthieuSchaller:
Does what is says on the tin.

Now, that's going to be a fun one for you. The data is given as 12 + log10(O/H) for historical reasons. I have get it that way (i.e. I did not undo the logarithm) as displaying it in any other format on a y-axis would be odd. This is the convention.

In these units, Z_sun = 8.69. (i.e. 12 + log10(O/H)_sun == 8.69)

SWIFT (and VR) returns metallicities as metal mass fractions. This means they are agnostic of Z_sun. 

So, once you assume a value of Z_sun (say 0.0126) then you can can express the simulation Z in units of Z_sun and match that to the data. 

You are welcome.


MatthieuSchaller:
I would recommend putting the value of Z_sun on any of the plots showing metallicity and also give the conversion from Z_sun to log(O/H).

MatthieuSchaller:
I have also added the Gallazzi+2005 stellar mass - stellar metallicity data.
That one is easier to deal with as somehow for stars people quote metallicities in units of solar.

JBorrow:
Hmm, why don't we convert the data to MMF and note the values used in the legend?

MatthieuSchaller:
MMF?

JBorrow:
metal mass fractions

JBorrow:
We can have this configurable through the `cosmology.py` file, so people can change the values globally later.

MatthieuSchaller:
You can but nobody will look at your plot as it won't have a "natural" metallicity axis.  A bit like if you were to show the stellar mass function in cgs.

JBorrow:
So we _should_ in fact be converting metallicities to solar?

MatthieuSchaller:
You must assume solar somewhere. Measurements are almost always done with respect to solar. 

So, you will have to either convert the data or the simulation. Now, there is about a factor 2 uncertainty on the value of Z_solar so it's important to specify what was used when going from mass fraction to Z (or vice versa). 

the thing is that the simulation does not need to assume anything about Z_sun ever. Which means we are immune to this uncertainty in the internals of the code. It's only a matter for comparison.

MatthieuSchaller:
I have added a second data set for the stellar mass - gas phase metallicity. Same units as the other one at least.

MatthieuSchaller:
One option is to add two axes like figure 13 of the original paper.

MatthieuSchaller:
Now I am done. :)

JBorrow:
Thanks!

Issue #21, opened 2020-04-01
Stellar Mass - Gas Z


MatthieuSchaller:
This could be useful to add for the low-mass end:

https://ui.adsabs.harvard.edu/abs/2006ApJ...647..970L/abstract

MatthieuSchaller:
Also 

https://ui.adsabs.harvard.edu/abs/2013ApJ...765..140A/abstract

PR #22, opened 2020-04-09
More mass metallicity

MatthieuSchaller:
Implements #21 

The Lee+2006 data will be good for COLIBRE. The Andrews+2013 is good for COLIBRE and XL.


JBorrow:
Thanks!

PR #29, opened 2020-07-06
Move the Kroupa --> Chabrier IMF correction out of the exp10() and fit

MatthieuSchaller:
I have now convinced myself that this is correct.


MatthieuSchaller:
All done on my side. Over to you.

PR #31, opened 2020-07-14
Add the z=0 EAGLE-100 Ref stellar mass - halo mass relation

MatthieuSchaller:
What do you think of the format? 

If that works, I'll add the 25Mpc box and other redshifts.


JBorrow:
Looks good, but can we change `.dat` for `.txt` or `.tsv`? `.dat` generally are binary files

MatthieuSchaller:
Also, can I use the same script to also produce M_200 - M_*/M_200 or do you want one separate script per dataset?

JBorrow:
Either is fine

JBorrow:
do you want to do that here or in a separate mr

MatthieuSchaller:
Working on it. Give me 1hr.

JBorrow:
no worries any time is fine just wanted to know whether to merge or not

MatthieuSchaller:
Done. :)

JBorrow:
So this is definitely a fit not a median line?

JBorrow:
Can you also change the comment to indicate how the relation has been reduced in those cases?

MatthieuSchaller:
It's the median line.

JBorrow:
Ok, we should change the description to state that instead of fit then

MatthieuSchaller:
Done.

JBorrow:
Sorry can you also change the name to indicate that they are the reduced versions in those two scripts

MatthieuSchaller:
What do you mean?

MatthieuSchaller:
The dedscription already says medians.

data/GalaxyStellarMassHaloMass/conversion/convertEAGLERatioStellar.py
```diff
@@ -0,0 +1,72 @@
...
+
+# Meta-data
+comment = (
+    "Medians obtained directly from the subfind catalogs. "
+    "Central galaxies only. Halos with 0 stellar mass included. "
+    "No cosmology correction needed."
+)
+citation = "EAGLE - L100N1504"
+bibcode = "2015MNRAS.446..521S"
+name = "Stellar mass - halo mass relation at z=0."
```
JBorrow, line 41: Change this to "Galaxy Stellar Mass / Halo Mass - Stellar Mass relation at z=0" for instance.

MatthieuSchaller:
Done.

PR #32, opened 2020-07-15
Update eagle smhm

MatthieuSchaller:
So, I am an idiot and did not make the correct ratio plots. I should bin M_*/M_200 and not just divide by the centre of the bin. 
That's not corrected.


MatthieuSchaller:
Anything missing here?

JBorrow:
It's better practice in the future to do `unit_array = unyt.unyt_array(numpy_array, units=unyt.this_unit)` than `unit_array = numpy_array * unyt.this_unit` but that's ok for now. thanks.

PR #33, opened 2020-07-15
Update behroozi smhm

MatthieuSchaller:
Update the Behroozi+2013 SM-HM data:
 - Use the data provided on their webpage instead of the fit of unknown origin.
 - Use the correct limits for the fit.
 - Apply a correction to get M_200 from the M_vir used in the paper.


data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2013.py
```diff
@@ -19,32 +18,37 @@
...
+name = "Fit to the stellar mass - halo mass relation at z=0."
 plot_as = "line"
 redshift = 0.0
 h = h_sim
 
 # Write everything
 processed = ObservationalData()
 processed.associate_x(
-    halo_masses * unyt.Solar_Mass, scatter=None, comoving=False, description="Halo Mass"
+    M_200, scatter=None, comoving=True, description="Halo Mass (M_200_cr)"
```
JBorrow, line 48: Please stylise with LaTeX "Halo Mass ($M_{200, {\rm crit}}$)"

MatthieuSchaller:
Done. Also for the other relations.

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2013.py
```diff
@@ -19,32 +18,37 @@
 if not os.path.exists(output_directory):
     os.mkdir(output_directory)
 
+data = np.loadtxt("../raw/Behroozi_2013_z0p1.txt")
```
chaykinea, line 21: @MatthieuSchaller based on the name of the file, on lines `40` and `42` it should probably be `z=0.1`

MatthieuSchaller: I guess you are formally right here. 

In practice the data that goes into this is  0 < z < 0.2 or something with complicated selection functions. Also at z < 0.5 there is little eveolution. 

But feel free to make the correction.

PR #34, opened 2020-09-13
Add hmf data

MatthieuSchaller:
That's it. Just halo mass functions.


data/HaloMassFunction/conversion/convertBocquet2016.py
```diff
@@ -0,0 +1,50 @@
...
+
+output_filename = "Bocquet2016.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+data = np.loadtxt("../raw/HMF_Colossus.txt")
+
+M_200 = data[:, 0] * unyt.Solar_Mass
```
JBorrow, line 21: These should be changed to be cast to `unyt` arrays as:

```python
M_200 = unyt.unyt_array(data[:, 0], units=unyt.Solar_Mass)
```

which avoids weird issues with conversions between units.

data/HaloMassFunction/conversion/convertBocquet2016.py
```diff
@@ -0,0 +1,50 @@
...
+Phi = data[:, 2] * unyt.Mpc ** (-3)
+
+comment = (
+    "Halo mass functions at z=0 from the colossus package (Diemer+18) assuming a 'planck13' cosmology. "
```
JBorrow, line 25: This line is too long; a maximum of 88 characters please.

JBorrow (changes requested):
Same comments for both

MatthieuSchaller:
Done.

PR #35, opened 2020-09-14
Add SFR functions

MatthieuSchaller:


data/StarFormationRateFunction/conversion/convertBell2007.py
```diff
@@ -0,0 +1,63 @@
...
+    exec(handle.read())
+
+output_filename = "Bell2007.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# The data is given in the form of a Schechter fit (Table 1 of the paper)
+phi_star = 9.8 * (10.0 ** -4) * unyt.Mpc ** (-3)  # Mpc^-3 dex^-1
```
JBorrow, line 19: Same comments here about casting to `unyt` arrays, instead of multiplying by units.

JBorrow (commented):
Other than the minor thing about casting the unyt arrays this looks good.

data/StarFormationRateFunction/conversion/convertBell2007.py
MatthieuSchaller: These are not arrays though. Is it also an issue for just scalars?

JBorrow: For scalars please cast them to `unyt_quantity`s

JBorrow: So `phi_star = unyt.unyt_quantity(9.8e-4, units="Mpc**-3")`

JBorrow: What you have works but it's more stable to use the cast to a quantity / array when using older versions of `unyt` (or possibly future versions where stuff changes...)

MatthieuSchaller:
Done.

PR #36, opened 2020-09-15
 …

MatthieuSchaller:
Use the REF EAGLE-25 SF history rather than the no-AGN model. Remove the EAGLE-12 box data that is unused and not representative anyway.


MatthieuSchaller:
Nice title github...

JBorrow:
Very ominous

PR #42, opened 2020-10-07
Normalise the gas fraction data to the cosmic mean

MatthieuSchaller:
We also need to normalize the data.

### DIFF
--- a/data/HaloMassGasFractions/conversion/convertLin2012.py
+++ b/data/HaloMassGasFractions/conversion/convertLin2012.py
@@ -11,6 +11,8 @@ with open(sys.argv[1], "r") as handle:
 
 # Cosmology
 h_sim = cosmology.h
+Omega_b = cosmology.Ob0
+Omega_m = cosmology.Om0
 
 input_filename = "../raw/Lin2012.dat"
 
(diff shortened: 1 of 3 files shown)


data/HaloMassGasFractions/conversion/convertLin2012.py
```diff
@@ -47,7 +53,7 @@
     "Ionized gas out of 94 clusters combining Chandra, WISE and 2MASS. "
     "Data was corrected for the simulation's cosmology."
 )
-citation = "Lin et al. (2012)"
+citation = "Lin et al. (2012) - (z < 0.25)"
```
JBorrow, line 56: This is not required, it will be added automatically by the pipeline.

MatthieuSchaller: No, it won't. This is using the z<0.25 data from Lin et al. i.e. I am trimming their catalogue of clusters to only use the low-z objects.

JBorrow: :/ I see. This is why I don't want to add the redshift labels automatically... Sigh. This will then have two redshift labels - it'll be shown as Lin et al. (2012) - (z < 0.25) (z = 0.2) or something...

MatthieuSchaller: We can remove it but then we need to specify it on the plot caption somehow as it actually makes an important difference.

JBorrow: Yes, you're right. Can we put this inside the year? Sorry for being nit picky, so - Lin et al. (2012; $z<0.25$ only)

JBorrow:
Great thanks

PR #45, opened 2020-10-12
Add Behroozi 2019 smhm

EvgeniiChaikin:
Adding Behroozi 2019 StellarMassHaloMass data.

**Note that**
- The three new scripts make use of the _velociraptor-library function_ `behroozi_2019_raw()`, which is currently not part of the master branch, but rather belongs to [this pull request](https://github.com/SWIFTSIM/velociraptor-python/pull/26). 
- The conversion factor of **1.2** in the new scripts is only valid for **z=0**. The conversion factor connects `M_vir` (what is passed to `behroozi_2019_raw()`) to `M_200` (what is plotted).
- As the next step, we need to agree on 
  1. What to do with the conversion factor for `z>0`. The suggestion was to make a separate plot(s) with `M_vir` in place of `M_200`. Then the conversion factor is no longer needed.
  2. The way to produce Behroozi 2019 high-z data in addition to z=0 data without having to edit some files and rerun `convert.sh` several times.
 @MatthieuSchaller  suggested inserting a loop over `z` in these three new scripts, which will make them generate multiple .hdf5 files (at various z) and require running `convert.sh` just once.
- I haven't done anything yet regarding the above two suggestions, because I don't want to complicate this PR too much before I receive any feedback.

### DIFF
--- a/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
+++ b/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
@@ -33,15 +33,15 @@ M_200 = M_vir / 1.2
 
 # Meta-data
 comment = (
-    "The data is taken from https://www.peterbehroozi.com/data.html"
-    "Median fit to the raw data for centrals (i.e. excluding satellites)"
-    "The stellar mass is the true stellar mass (i.e. w/o observational corrections)"
-    "The halo mass is the peak halo mass that follows the Bryan & Norman (1998)"
-    "spherical overdensity definition"
-    "The fitting function does not include the intrahalo light contribution to the"
-    "stellar mass"
-    "Cosmology: Omega_m=0.307, Omega_lambda=0.693, h=0.678, sigma_8=0.823, n_s=0.96"
-    "Shows the stellar mass as a function halo mass"
+    "The data is taken from https://www.peterbehroozi.com/data.html. "
+    "Median fit to the raw data for centrals (i.e. excluding satellites). "
+    "The stellar mass is the true stellar mass (i.e. w/o observational corrections). "
+    "The halo mass is the peak halo mass that follows the Bryan & Norman (1998) "
+    "spherical overdensity definition. "
+    "The fitting function does not include the intrahalo light contribution to the "
+    "stellar mass. "
+    "Cosmology: Omega_m=0.307, Omega_lambda=0.693, h=0.678, sigma_8=0.823, n_s=0.96. "
+    "Shows the stellar mass as a function halo mass."
 )
 citation = "Behroozi et al. (2019)"
 bibcode = "2019MNRAS.488.3143B"
--- a/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
+++ b/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
@@ -13,66 +13,68 @@ with open(sys.argv[1], "r") as handle:
 # Cosmology
 h_sim = cosmology.h
 
-output_filename = "Behroozi2019.hdf5"
-output_directory = "../"
+# Redshifts at which to plot the data
+redshifts = [0.0, 0.2, 0.5, 1.0, 2.0, 3.0, 4.0, 5.0]
 
-if not os.path.exists(output_directory):
-    os.mkdir(output_directory)
+# Create and save data for each z in redshifts
+for z in redshifts:
 
-# Redshift at which to plot the data
-redshift = 0.0
+    output_filename = f"Behroozi2019_{f'z{z:.1f}'.replace('.', 'p')}.hdf5"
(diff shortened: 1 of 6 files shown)


JBorrow:
Hi, this is failing the tests. Please check why (click `Details`) and resolve it.

EvgeniiChaikin:
Hi Josh, I indicated in the description why it is failing the tests. Namely, the scripts call a function that is not part of the velociraptor-library's master branch yet.

EvgeniiChaikin:
@JBorrow With your latest updates to the velociraptor library,  all tests have now been passed. Let me know if there is anything else that needs to be done in this PR.

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
```diff
@@ -0,0 +1,78 @@
...
+redshift = 0.0
+
+M_vir = np.logspace(9, 15, 512)
+
+# Note that the function below actually takes M_{vir, peak}, and we are
+# ignoring this fact
+M_star = behroozi_2019_raw(redshift, M_vir)
+
+# Correct M_vir --> M_200 (fit from EAGLE cosmology - cosmology dependence is very weak)
+M_200 = M_vir / 1.2
```
JBorrow, line 32: Do not do this! We should just make sure to plot this against `R_BN98` from the VR catalogues.

EvgeniiChaikin: yes, I am aware of that.

My original plan was to make the new data available ASAP (and hence temporarily put the data on the M_200 plot)

And next, as part of another PR, push the changes that would create a M_vir plot for the Berhoozi data.

JBorrow: I see. I don't want to merge data that we _know_ is incorrect. If you change this, then adding the plot for M_* against M_BN98 is basically a one line change in the pipeline config.

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
```diff
@@ -13,66 +13,68 @@
...
+redshifts = [0.0, 0.2, 0.5, 1.0, 2.0, 3.0, 4.0, 5.0]
 
-if not os.path.exists(output_directory):
-    os.mkdir(output_directory)
+# Create and save data for each z in redshifts
+for z in redshifts:
 
-# Redshift at which to plot the data
-redshift = 0.0
+    output_filename = f"Behroozi2019_{f'z{z:.1f}'.replace('.', 'p')}.hdf5"
```
JBorrow, line 22: Please use consistent formatting with the other files:

```
def stringify_z(z):
    """
    Eagle-style text formatting of redshift label.
    Example: z=1.5 will be printed as z001p500.
    z: The redshift to produce a label for
    """
    whole = int(z)
    frac = int(1000 * (z - whole))
    return f"z{whole:03d}p{frac:03d}"
```
You will probably be able to do a better job at this than that function, but we should keep things consistent.

EvgeniiChaikin: Oh, I thought the format was `z?p?`, not  `z???p???`

Because otherwise,

`https://github.com/SWIFTSIM/velociraptor-comparison-data/blob/master/data/GalaxyStellarMassHaloMass/conversion/convertEAGLE.py`

is already inconsistent.

Btw, can I import `stringify_z(z) ` from somewhere or do I need to define it in each file?

EvgeniiChaikin: Also, I think in many files we should change

`description="Halo Mass ($M_{200, {\rm crit}}$)" -> description=`**r**`"Halo Mass ($M_{200, {\rm crit}}$)"`

If the string is not a raw string, then in the produced .hdf5 files we will see `m` in place of `\rm`.

JBorrow: No it's zXXXpYYYY. I think you can do this with "z{:07.3f}".format(x).replace(".", "p"), though.

JBorrow: on the string - sure. I think it's best to do `"Halo Mass ($M_{200, {\\rm crit}}$)"` instead of making it a raw string though

EvgeniiChaikin: Done

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
EvgeniiChaikin: I've updated that. Now it is M_BN98

JBorrow:
Thanks!

PR #47, opened 2020-10-21
Fixed errorbars and redshift error

JBorrow:
Fixes the errorbar and redshift error in the Muzzin 2013 data.


JBorrow:
stellar_mass_evolution
With the fixed data

MatthieuSchaller:
I see what I did wrong here.

The thing here is that assigning a single z here makes little sense.

JBorrow:
Indeed - but I don't think it's so bad as long as we know that the value in the legend is the mean redshift of the observations.

MatthieuSchaller:
Good for me

JBorrow:
Let's wait on the CI to see if I messed anything up and then I'll merge.

PR #49, opened 2020-10-27
Add data from Lovisari et al (2015) and Eckert et al (2016) 

Moyoxkit:
Added extra data to the halo-mass gas fraction plots, at the moment these do not have error bars.


data/HaloMassGasFractions/conversion/convertEckert2016.py
```diff
@@ -0,0 +1,62 @@
...
+output_filename = "Eckert2016.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Read the data
+raw = np.loadtxt(input_filename)
+M_500 = unyt.unyt_array((0.70 / h_sim) * 10 ** raw[:, 0], units="Msun")
+fb_500 = unyt.unyt_array( raw[:, 1] * (0.70 / h_sim) ** (2.5)) / (Omega_b / Omega_m)
```
JBorrow, line 28: What are the units of this array?

data/HaloMassGasFractions/conversion/convertEckert2016.py
```diff
@@ -0,0 +1,62 @@
...
+
+# Meta-data
+comment = (
+    "Based on observations obtained with XMM-Newton. "
+    "Data was corrected for the simulation's cosmology."
```
JBorrow, line 33: Not necessarily true - it's corrected for the cosmology in `cosmology.py`. Might be worth printing that cosmology name.

data/HaloMassGasFractions/conversion/convertLovisari2015.py
```diff
@@ -0,0 +1,61 @@
...
+output_filename = "Lovisari2015.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Read the data
+raw = np.loadtxt(input_filename)
+M_500 = unyt.unyt_array((0.70 / h_sim) * 10 ** raw[:, 0], units="Msun")
+fb_500 = unyt.unyt_array( raw[:, 1] * (0.70 / h_sim) ** (2.5)) / (Omega_b / Omega_m)
```
JBorrow, line 28: Same comment as above.

data/HaloMassGasFractions/conversion/convertLovisari2015.py
```diff
@@ -0,0 +1,61 @@
...
+
+# Meta-data
+comment = (
+    "Data was corrected for the simulation's cosmology."
```
JBorrow, line 32: Comment here on h-factors, and again the cosmology name.

JBorrow (commented):
Just a couple of comments on your metadata. Please also ensure your python files are formatted with `black`, or they won't pass the tests.

data/HaloMassGasFractions/conversion/convertEckert2016.py
Moyoxkit: This array is dimensionless, it is a fraction

JBorrow: Sorry, I should have been clearer - I mean you should have `units="dimensionless"` as a parameter to that call to `unyt_array` :).

Moyoxkit:
I made some changes to the metadata and ran the formatting script. I assumed the notes about the cosmology were correct as I just copied them from another covert file

PR #50, opened 2020-10-28
Improved eagle Mstar - Mhalo

MatthieuSchaller:
Add more redshifts and halo mass definitions.


MatthieuSchaller:
@EvgeniiChaikin you will be interesting in this.

PR #53, opened 2020-11-18
Convert Muzzin 2013 data to use multi-redshift format

JBorrow:
Converts the Muzzin 2013 GSMF data to use the new multi-redshift data files.


JBorrow:
Requires https://github.com/SWIFTSIM/velociraptor-python/pull/38/files

PR #54, opened 2020-12-02
Make Berhoozi2019 & Eagle2015 data work with multi-z update

EvgeniiChaikin:
We already have **Berhoozi2019** and **Eagle2015** available at various redshifts.

However, following @JBorrow latest update to VR library, these data now needs to be arranged in an hdf5 file in a different way (so that VR can access the data at different redshifts **automatically**)

This is what this PR is exactly for.

### DIFF
--- a/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
+++ b/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
@@ -66,7 +66,7 @@ for z in redshifts:
     # Stellar masses (for the given halo masses, at redshift z)
     M_star = behroozi_2019_raw(z, M_BN98)
 
-    # A fitting function gives us the data at any z. Hence, no need to have \Delta z
+    # Choose \Delta z = 0.25 to cover the whole range in z
     redshift_lower, redshift_upper = [z - 0.25, z + 0.25]
 
     processed.associate_x(
--- a/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
+++ b/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
@@ -50,6 +50,7 @@ multi_z.associate_citation(citation, bibcode)
 multi_z.associate_name(name)
 multi_z.associate_comment(comment)
 multi_z.associate_cosmology(cosmology)
+multi_z.associate_maximum_number_of_returns(1)
 
 output_filename = "Behroozi2019.hdf5"
 output_directory = "../"
--- a/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
+++ b/data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
@@ -1,4 +1,7 @@
-from velociraptor.observations.objects import ObservationalData
+from velociraptor.observations.objects import (
+    ObservationalData,
+    MultiRedshiftObservationalData,
+)
 from velociraptor.fitting_formulae.smhmr import behroozi_2019_raw
 
 import unyt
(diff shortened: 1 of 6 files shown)


MatthieuSchaller:
What would you like me to check specifically?

EvgeniiChaikin:
@MatthieuSchaller I'd like to know If you are fine with how I modified the EAGLE2015-data convert files that you added some time ago.

MatthieuSchaller:
Seems all fine by me but aren't the z low and z upper a bit too narrow for the EAGLE data? Surely we want them on even if there is a slight mismatch in z.

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
```diff
@@ -10,49 +13,62 @@
...
-        "n_s=0.96. "
-        "Shows the stellar mass as a function halo mass."
-    )
-    citation = "Behroozi et al. (2019)"
-    bibcode = "2019MNRAS.488.3143B"
-    name = "Fit to the stellar mass - halo mass relation at z={:.1f}".format(z)
-    plot_as = "line"
-    h = h_sim
+    # A fitting function gives us the data at any z. Hence, no need to have \Delta z
+    redshift_lower, redshift_upper = [z, z]
```
JBorrow, line 70: A couple of comments on this specific point.

I agree with Matthieu that we want to extend this to the range 'covered', so e.g. let's do 0.0 to 0.2 for the first. The SMHMR should then be calculated at the mean redshift, or maybe the mean scale factor.

data/GalaxyStellarMassHaloMass/conversion/convertEAGLE.py
```diff
@@ -41,23 +73,9 @@
...
-        bibcode = "2015MNRAS.446..521S"
-        name = f"Galaxy Stellar Mass - Halo Mass relation at z={f'{z:3.1f}'}."
-        plot_as = "line"
-        redshift = z
-        h = h_sim
-
-        # Write everything
-        processed = ObservationalData()
+        # A fitting function gives us the data at any z. Hence, no need to have \Delta z
+        redshift_lower, redshift_upper = [z, z]
```
JBorrow, line 77: I suppose for EAGLE where it is calculated at exactly z=z we can probably leave this and let the auto bracketing deal with things.

JBorrow:
Just a couple of minor comments on the redshifts. The rest looks good to me. Thanks for being patient and waiting for this review.

data/GalaxyStellarMassHaloMass/conversion/convertBehroozi2019.py
EvgeniiChaikin: How about we use `z \pm 0.2`? And the mean redshift is just `z`?

JBorrow: In that case no relation would be plotted for a z=3.5 or z=4.5 snapshot.

JBorrow: (It would, technically, but let's not rely on parameters in the rest of the codebase saving the day when we can avoid it).

EvgeniiChaikin: I see. I just update the PR. Berhoozi2019 data should now span the whole range in z

JBorrow: If we are going to go with a fixed offset why not do:
[0, 0.125] -> calc at 0
[0.125, 0.375] -> calc at 0.25
[0.375, 0.625] -> calc at 0.5
...
all the way up. It's not like there's gonna be an insane amount of data there.

EvgeniiChaikin: Done. But now there is more code, which I wanted to keep as simple as possible

PR #55, opened 2021-01-04
Add new data for dust property comparisions

wmfw23:
including data from Rémy-Ruyer et al I2014) and DeLooze et al (2020).

Running black also reformatted the De Souza et al  (2015) GSMF


wmfw23:
@JBorrow so it seems like the build here fails because I tried importing `scipy`, which was necessary to convert the raw  De Looze data (individual galaxies) to binned medians. Can we include  `scipy` as a dependency or do we need to find another way?

JBorrow:
Does `binned_median_line` from velociraptor-python work for you? https://velociraptor-python.readthedocs.io/en/latest/modules/velociraptor.tools.lines.html#velociraptor.tools.lines.binned_median_line

JBorrow:
Thanks!

PR #57, opened 2021-01-28
Add scatter to Behroozi2019 data and move it into a single script

EvgeniiChaikin:
Hi @JBorrow 

In this PR, I

- Move all Behroozi 2019 scripts into a single file.

- Add errorbars to all Behroozi2019 data

**Example of the error bars:**

_Before:_
stellar_mass_halo_mass_MBN98_centrals_ratio_100_2

_After:_
stellar_mass_halo_mass_MBN98_centrals_ratio_100


JBorrow:
Similar to the passive fractions, I'm not convinced by the highly asymmetric errors here. The code looks good though.

EvgeniiChaikin:
Yeah, that looks a bit weird. But that's what Behroozi provides. See e.g. https://arxiv.org/abs/1806.07893, the left panels of Fig. 9 and 10.

JBorrow:
Okay, happy with this then. Thanks!

PR #63, opened 2021-02-04
Add Reed2017 data

### LINKED ISSUE #59: Add Read+2017 M*-MH data
https://ui.adsabs.harvard.edu/abs/2017MNRAS.467.2019R/abstract

EvgeniiChaikin:
Hi @JBorrow 

In this PR, I add stellar vs halo mass data from Reed et al (2017)

- There are 11 field dward galaxies that have well measured rotation curves and hence are present in the plots
- For the redshift inverval I chose the values [-0.1 and 2.1] becasue we also want this data at high z.
- The original paper actually gives more than 11 galaxies, but the remaining ones either have bad data or no rotation curves at all
- The first plot  below should be compared with Fig. 3 in https://ui.adsabs.harvard.edu/abs/2017MNRAS.467.2019R/abstract

Any feedback is welcome!

<h2>Example of new SMHM plots</h2>

stellar_mass_halo_mass_M200_all_100
stellar_mass_halo_mass_M200_centrals_ratio_100
stellar_mass_halo_mass_M200_centrals_ratio_stellar_100

Fixes #59


JBorrow:
The structure of functions that use global variables is a bit concerning to me. If we can remove the function component and just have them done one after each other in the script, I think that would make more sense.

JBorrow:
Excellent, thanks for the contribution!

PR #72, opened 2021-02-22
Add z=4-7 GSMF from Duncan et al 2014

calvin-sykes:


data/GalaxyStellarMassFunction/conversion/convertDuncan2014.py
```diff
@@ -0,0 +1,147 @@
...
+citation = "Duncan et al. (2014)"
+bibcode = "2014MNRAS.444.2960D"
+name = "GSMF from CANDELS/GOODS-S"
+
+multi_z = MultiRedshiftObservationalData()
+multi_z.associate_comment(comment)
+multi_z.associate_name(name)
+multi_z.associate_citation(citation, bibcode)
+multi_z.associate_cosmology(cosmology)
+
```
JBorrow, line 136: Do you want to add here the https://velociraptor-python.readthedocs.io/en/latest/modules/velociraptor.observations.objects.html#velociraptor.observations.objects.MultiRedshiftObservationalData.associate_maximum_number_of_returns value?

calvin-sykes: So would this just be 1? (on the basis that if you're plotting a snapshot at a given z, you'd just want the nearest observed z bin to compare with)

JBorrow:
Yeah, if you don't set it it will return all data that overlap with the requested redshift range

PR #73, opened 2021-02-22
Add z=3.5-7.5 GSMF from Grazian et al 2014 

calvin-sykes:


JBorrow:
Same comment as your other MF, let me know your thoughts here when you're ready to merge.

PR #74, opened 2021-02-22
Add z=3.8-6.8 GSMF from Gonzalez et al 2011

calvin-sykes:
Found another one :laughing: 

I promise this is the last one though - now all the GSMFs from the eagle paper (and Furlong+15) are available


JBorrow:
Thanks for all these!

PR #75, opened 2021-02-23
Composite Neutral Gas Data

wmfw23:
Neutral gas fraction data by combining Catinella+18 and Saintonge+17 data sets.


JBorrow:
I've assigned @EvgeniiChaikin to give this a draft review as he'll be taking over the maintenance of the repo once I leave for MIT

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
...
+    "Galaxy Specific Star Formation Rate",
+]
+
+filetag = ["abcissa_M_star", "abcissa_mu_star", "abcissa_sSFR"]
+
+
+for i in range(len(tables)):
+    processed = ObservationalData()
+
+    if i < 3:
```
EvgeniiChaikin, line 69: Can you please add a few short comments to this section explaining what is going on here? Otherwise, it is difficult to follow

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
+from velociraptor.observations.objects import ObservationalData
+
+import unyt
+import numpy as np
+import os
+import re
+import sys
+import itertools as it
```
EvgeniiChaikin, line 8: `intertools` and `re` are not used in this file (and the same is also true for convertCatinellaSaintongeComposite.py)

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
...
+
+ORIGINAL_H = 0.7
```
EvgeniiChaikin, line 10: If this value of h was used in Catinella2018, can you please rename this variable it to `h_Catinella2018` or add a comment above that the value of 0.7 comes from this paper?

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
...
+for i in range(len(tables)):
+    processed = ObservationalData()
+
+    if i < 3:
+        x_vals = 10 ** tables[i][:, 0] * units[i]
+    else:
+        x_vals = tables[i][:, 0] * units[i]
+
+    # no x err
+    x_err = x_vals * 0.0
```
EvgeniiChaikin, line 75: If the data provides no scatter it is probably better to set `scatter=None` (in function `processed.associate_x/y`) instead of defining arrays with zero errors

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
...
+
+    fgas = 10 ** tables[i][:, 1] * unitless
+
+    fgas_err = fgas * 0.0
+
+    processed.associate_x(x_vals, scatter=x_err, comoving=0, description=labels[i])
+    processed.associate_y(
+        fgas,
+        scatter=fgas_err,
+        comoving=0,
```
EvgeniiChaikin, line 85: `comoving` is a `bool`; please use `True` of `False` for clarity

EvgeniiChaikin (changes requested):
Hi @wmfw23 I finished my review and sorry for the delay.

I left a number of comments here and there whose main goal is to improve the readability of the files.

Note that the comments I wrote in `convertCatinella2018.py` are also relevant for `convertCatinellaSaintongeComposite.py` 

Also, can you please show a few examples from the pipeline with these new data on the plots? I want to make sure everything is displayed and behaves correctly.

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
```diff
@@ -0,0 +1,100 @@
...
+
+readrows = np.hstack([np.diff(table_index).astype(int), None])
+
+tables = []
+for i in range(len(table_index)):
+    tables.append(
+        np.loadtxt(input_filename, skiprows=table_index[i], max_rows=readrows[i])
+    )
+
+citation = "Catinella et al. (2018), z = 0"
```
EvgeniiChaikin, line 43: Ah, sorry, forgot to say -  there is no need to indicate redshift in the citation because otherwise it will appear twice in the plot

See, e.g.,

h2_frac_func_stellar_mass

data/GalaxyHItoStellarFractionMetallicity/convertDeLooze20.py
```diff
@@ -0,0 +1,168 @@
...
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+lines = []
+labels = []
+inc = 0
+
+# find all the sub headers in the file (tthey contain a ':')
```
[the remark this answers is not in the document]
wmfw23: this file doesn't actually do anything and was added by mistake as a copy of an already accepted data script, so have removed and modified and added the correct file

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
wmfw23: this isn't relevant anymore as don't include the colours in the raw file, so simplified it.

data/GalaxyColdGasFractions/conversion/convertCatinella2018.py
wmfw23: this isn't relevant anymore as don't include the colours in the raw file, so simplified it.

EvgeniiChaikin: Thanks!

EvgeniiChaikin:
Hi @wmfw23,

I have had a brief second look at the changes. 

Just a few tiny remaining comments, and then the branch can be merged:

- There are still a few places in the code where you are using `comoving=0` instead of `comoving=True/False ` and `scatter=array with zeros` instead of `scatter=None`
-  I don't understand why in `convertDeLooze20.py` you define `inc=0` and increment it by one by hand at the end of each iteration (`enumerate` is doing it for you anyway)
- Can you please add the same comment "_From cosmology assumed by Catinella+18_" to `convertCatinellaSaintongeComposite.py`?

EvgeniiChaikin (approved):

EvgeniiChaikin:
Thank you @wmfw23! I think this all looks very good now!

@JBorrow if you were not planning to have a final look, I am ready to merge it.

data/GalaxyHItoStellarFractionMetallicity/conversion/convertDeLooze20.py
```diff
@@ -24,24 +23,24 @@
 labels = []
 inc = 0
```
JBorrow, line 24: `inc` doesn't need to be initialised here

JBorrow:
Other than my one nit pick we're good to go.

PR #80, opened 2021-03-25
add new low mass data for better small-box comparison

wmfw23:
lots of new data for comparing gas properties, particularly @ low mass


data/CosmicH2Abundance/raw/#Peroux2020_OmegaH2.txt#
```diff
@@ -0,0 +1 @@
+# redshift | log Omega_H2≈
```
JBorrow, line 1: Do we need this empty file? What's with the hashes in the filename?

JBorrow (commented):
Only minor comment is about the weird file, thanks for the contribution!

PR #86, opened 2021-06-11
Add BH masses from Marasco+21

ymbahe:
Add the M_BH -- M_halo data from Marasco+21 to the observational comparison data set. Note that I have not been able to run the formatting script, so let me know if there are any remaining formatting issues.


JBorrow:
Can you add the comment from the top of the raw file to the comment in the HDF5?

ymbahe:
Sure, will do. I also need to fix an issue with using the data in the pipeline, hang on...

ymbahe:
With the last commit, the data now actually show up properly in the pipeline.

PR #88, opened 2021-06-25
Added GSMF data from Leja+ (2020).

jemme07:
Added GSMF data from Leja+ (2020). Uses panchromatic SED models that infer systematically higher masses and lower star formation rates than standard approaches.

### DIFF
--- a/data/GalaxyStellarMassFunction/conversion/convertLeja2020.py
+++ b/data/GalaxyStellarMassFunction/conversion/convertLeja2020.py
@@ -12,8 +12,8 @@ z_list = [0.2, 0.5, 1, 1.5, 2, 3]
 
 output_filename = "Leja_2020.hdf5"
 output_directory = "../"
-comment = f"Uses panchromatic SED models that infer systematically higher masses and lower star formation rates than standard approaches. h-corrected for SWIFT using cosmology: {cosmology.name}."
-citation = "Leja et al. 2020"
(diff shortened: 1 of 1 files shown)


MatthieuSchaller:
Does it require any h-correction or are they doing things in h-free coordinates? 

Can we specify the IMF choice in the description?

jemme07:
Everything is in h-free coordinates.

MatthieuSchaller:
But did they assume the same h as us when constructing their results?

jemme07:
Just added the IMF info. 

Right, regarding conversion , they used WMAP9 and give everything in h-free. So, if we would like to change cosmology, we would need to convert to h-units and back to h-free using the appropriate h-factors. This should be a very a small effect as the h values are now all very close to 0.7. I can include the conversions if you think is necessary.

MatthieuSchaller:
We have typically done that for the other datasets. Factots of h^3 (as you'd expect on the y-axis) lead to a 10% difference between h=0.7 and EAGLE h.

jemme07:
No worries, I'll do that.

jemme07:
Done! Added the cosmology correction factors.

MatthieuSchaller:
Thanks.

Sorry to be a pain, are we sure the masses are Msun/h and not Msun/h^2? That can be the case in SED fitting models.

jemme07:
I did h^2, that should be the masses inferred from from luminosities.

MatthieuSchaller (approved):

PR #91, opened 2021-07-04
Add the Driver+2021 GAMA DR4 galaxy mass function data

MatthieuSchaller:
Does what it says on the tin. 

Will correct all the relevant fields when eventually published.


Moyoxkit:
Looks good!

MatthieuSchaller:
Do you agree we should apply the correction they quote in the table header?

Moyoxkit:
Yes, this reduced the error still left due to cosmic variance by a lot, and allows us to use the 6.5% quoted at the end of the paper.

PR #92, opened 2021-07-06
Add MHI/Mstar data from local dwarf galaxies

james-trayford:


JBorrow:
Can you convert from Kroupa to Chabrier? We provide a conversion factor in `cosmology.py` that you can use. Please also note it in the comment.

JBorrow:
Can you include the value in the comments so we can refer back to it easily? I always feel more comfortable with metadata duplicated. Thanks!

Also just checking as the CI is down that this is all black formatted?

MatthieuSchaller:
What is the status of this?

PR #93, opened 2021-07-13
Adding stellar abundance raw data

correac:
Adding some data points for the stellar abundances plot. My own compilation as well as input from Rob Crain a year or so ago.


JBorrow:
Do you have the associated conversion files? That way we can keep track of the citations, any weird things about the files, and so on. 

Loading raw text in the pipeline (instead of loading observational data) is highly discouraged.

correac:
I don't have conversion, only raw.

bwvdnbro:
@correac could you please clarify which paper the Koch (2008) observations come from? There are multiple papers that match this reference and I was unable to determine which one is the source of the raw data you added.

bwvdnbro:
@correac could you confirm that Mishenina et al. (1999) is in fact Mishenina et al. (2000) (https://ui.adsabs.harvard.edu/abs/2000A%26A...353..978M/abstract)?

bwvdnbro:
@correac: I have opened a pull request for the conversion scripts in your fork. I currently don't have write access to this repository yet.

correac:
Hi Bert @bwvdnbro , thank you so much for your help on this. The Koch+08 paper refers to (https://ui.adsabs.harvard.edu/abs/2008AJ....135.1580K/abstract). Correct, Mishenina+1999 is the reference you provided.

PR #97, opened 2021-07-16
Fixed CI errors.

bwvdnbro:
Fixed two errors flagged by CI:

1. #96: conversion scripts that failed on rerun
2. Scripts that had not been properly formatted using `black`

Issue #96, opened 2021-07-16
Duncan (2014) and Gonzalez (2011) stellar mass function conversion fails upon second run


bwvdnbro:
The conversion scripts `data/GalaxyStellarMassFunction/conversion/convertDuncan2014.py` and `data/GalaxyStellarMassFunction/conversion/convertGonzalez2011.py` do not remove previously created output files and as a result fail upon a rerun of `convert.sh`.

PR #98, opened 2021-07-27
First version of CI

bwvdnbro:
This pull request adds CI to the repository (again?) using GitHub Actions. The new action workflow first installs the Python module dependencies (currently `velociraptor` and `matplotlib`) and then runs the conversion script twice. The second run is necessary to check if all conversion scripts properly clean up old output files. As an example why this is useful, the current `master` does not pass the test for exactly that reason (see #96 and #97).
The CI runs for every push to `master` and every pull request for `master`. To test it, it currently runs for an artificial pull request in my own fork, see https://github.com/bwvdnbro/velociraptor-comparison-data/actions/runs/1071776187.

Todo:
- [x] Add formatting check

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,8 +31,8 @@ jobs:
 
       # Runs the conversion script for the first time
       - name: First conversion
-        run: ./convert.sh
+        run: bash convert.sh
         
       # Runs the conversion script again to check proper output file cleanup
       - name: Second conversion
-        run: ./convert.sh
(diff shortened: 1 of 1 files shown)


bwvdnbro:
All checks seem to work as intended now. Both fail for the current `master`.

JBorrow:
Are you happy to fix the issues in a separate MR?

bwvdnbro:
Yes. Well, #97 already fixes one of them. The other one is simply a matter of running the format script.

PR #99, opened 2021-07-28
Added X-ray tables and hot gas data from BAHAMAS (and other observational datasets).

jemme07:
Added X-ray tables (not sure if this has to be here). Also added (unpolished) hot gas data from BAHAMAS and other observational datasets.


bwvdnbro:
@jemme07 Which files are the X-ray tables? To me, it looks like all these files are observational data or BAHAMAS data.

Before this can be merged, we need
 - [ ] a conversion script for all of these data files, so that the data handling can be moved from the pipeline to this repository, which is the whole point of having separate repositories,
 - [ ] a full reference for all the data, especially the BAHAMAS data that has no information right now.

I'm happy to help with the conversion scripts, but then you will need to at least provide me with the full references.

jemme07:
Thanks @bwvdnbro!

Here is the link to the BAHAMAS paper with all the reference information:

https://ui.adsabs.harvard.edu/abs/2017MNRAS.465.2936M/abstract

bwvdnbro:
@jemme07 Thanks for that reference, that helps a lot.

I am still missing a full reference for the Rasmussen data though. Could you tell me where you got these data?

MatthieuSchaller:
@jemme07 any updates on that last request?

jemme07:
Here is the link to the Rasmussen paper:

https://ui.adsabs.harvard.edu/abs/2007MNRAS.380.1554R/abstract

Please let me know if you need any additional information.

bwvdnbro:
@jemme07 This should be good to go, if you can merge the changes made in https://github.com/jemme07/velociraptor-comparison-data/pull/1 into your fork.

MatthieuSchaller:
This has fallen between the cracks; sorry.

It looks like there are some test issues. @jemme07 could you fix the formatting and push once more?

PR #100, opened 2021-08-30
add SFH from Behroozi 2019 

EvgeniiChaikin:
**add SFH from Behroozi 2019** 

**Paper:** https://arxiv.org/pdf/1806.07893.pdf

**Data:** (from the paper above, left plot)

Screenshot from 2021-08-30 14-15-05


JBorrow:
All data (including SFH data) should now come with a conversion script. I think the inclusion of data without conversion scripts (and hence significant missing metadata) is a mistake that we don't want to repeat.

EvgeniiChaikin:
Yes, of course! But currently, the star formation history plot is an exception to this rule. See this: https://github.com/SWIFTSIM/pipeline-configs/blob/master/colibre/scripts/load_sfh_data.py

I agree that ideally, we need to properly (re-)write conversion scripts for each data sets that are present in that plot.

JBorrow:
Yes, I'm aware that we have lots of data that doesn't come along with the metadata for SFH, but when we add new data there's no reason to not 'do it properly', even if the legacy data is still wrong.

JBorrow:
I will let @bwvdnbro take it from here. Thanks Evgenii for your fast turnaround!

JBorrow:
There are minor differences between black versions. Can you try updating the one you're using to the latest release?

EvgeniiChaikin:
Thanks @JBorrow 

Yes, indeed, doing

`python3.8 -m pip install black --upgrade`

solved the problem

JBorrow:
Ok, thanks! I'll let somebody else make the final merge decision as I've been away for a while and I'm not sure if anythings changed but I'm happy with this as-is

MatthieuSchaller (approved):

PR #102, opened 2021-11-03
Add the Driver+2021 stellar mass density point

MatthieuSchaller:
To be consistent with the GSMF we now use.


MatthieuSchaller:
Two good points. I'll get to it.

MatthieuSchaller:
How did you get 0.615? 

I get 2.17 * (0.04 + 0.58 + 0.2) / 2.97 == 0.599

Moyoxkit:
I did something stupid, your number is correct.

PR #103, opened 2021-11-12
Adding gallazzi data

correac:
Adding a stellar mass-[alpha/Fe] relation taken from Gallazzi+ (2021) (https://ui.adsabs.harvard.edu/abs/2021MNRAS.502.4457G/abstract)

### DIFF
--- a/data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
+++ b/data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
@@ -25,7 +25,7 @@ if not os.path.exists(output_directory):
 
 # Read the data
 raw = np.loadtxt(input_filename, delimiter=delimiter)
-M_star = 10 ** raw[:, 0] * unyt.Solar_Mass * (h_sim / h_obs) ** -2
+M_star = 10 ** raw[:, 0] * unyt.Solar_Mass * (h_obs / 0.7) ** 2
 
 # Correction factor due to the difference in (X_O/X_Fe)_Sun
 # from Grevesse & Sauval (1993) to Asplund+ (2009)
@@ -56,7 +56,9 @@ y_scatter = unyt.unyt_array((Z_median - Z_lo, Z_hi - Z_median))
 comment = (
     "Data obtained assuming a Chabrier IMF and h=0.7. "
     f"h-corrected for SWIFT using cosmology: {cosmology.name}. "
-    "The metallicity is expressed as [alpha/Fe]. "
+    "The metallicity is expressed as [alpha/Fe]. Note that alpha does not stand for Oxygen. "
+    "Gallazi et al. adopt a semi-empirical estimate of [alpha/Fe] building on the work of Gallazzi et al. (2006). "
+    "For each galaxy they measure the index ratio Mgb/Fe. "
     "The error bars given the 16th and 84th percentile of the distribution. "
     f"This has been corrected to use Z_solar={solar_metallicity} (Asplund+ 2009)"
 )
--- a/data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
+++ b/data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
@@ -1,0 +1,90 @@
+from velociraptor.observations.objects import ObservationalData
+
+import unyt
+import numpy as np
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+# Cosmology
+h_sim = cosmology.h
+h_obs = 0.704  # WMAP7
+Z_solar_obs = 0.02
+
+input_filename = "../raw/Gallazzi_2021_ascii.txt"
+delimiter = "\t"
+
+output_filename = "Gallazzi2021_Data.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Read the data
+raw = np.loadtxt(input_filename, delimiter=delimiter)
+M_star = 10 ** raw[:, 0] * unyt.Solar_Mass * (h_sim / h_obs) ** -2
+
+# Correction factor due to the difference in (X_O/X_Fe)_Sun
+# from Grevesse & Sauval (1993) to Asplund+ (2009)
+
+O_over_H_Grevesse93 = 8.83  # Grevesse & Sauval
+Fe_over_H_Grevesse93 = 7.5  # Grevesse & Sauval
+
+O_over_H_Andres89 = 8.93
+Fe_over_H_Andres89 = 7.51
+
+O_over_H_Asplund09 = 8.69
+Fe_over_H_Asplund09 = 7.50
+
+O_over_Fe_solar_Grevesse93 = O_over_H_Grevesse93 - Fe_over_H_Grevesse93
+O_over_Fe_solar_Andres89 = O_over_H_Andres89 - Fe_over_H_Andres89
+O_over_Fe_solar_Asplund09 = O_over_H_Asplund09 - Fe_over_H_Asplund09
+
+correction_Sun_O_over_Fe = O_over_Fe_solar_Grevesse93 - O_over_Fe_solar_Asplund09
+
+Z_median = (raw[:, 1] + correction_Sun_O_over_Fe) * unyt.dimensionless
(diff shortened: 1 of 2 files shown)


data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
```diff
@@ -0,0 +1,90 @@
...
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+# Cosmology
+h_sim = cosmology.h
+h_obs = 0.704  # WMAP7
+Z_solar_obs = 0.02
+
+input_filename = "../raw/Gallazzi_2021_ascii.txt"
```
bwvdnbro, line 17:
```suggestion
input_filename = "../raw/gallazzi_2021_ascii.txt"
```
As also revealed by the automatic testing: the name used here is not the same as the name of the actual file.

bwvdnbro (changes requested):

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
```diff
@@ -0,0 +1,90 @@
...
+Z_hi = (raw[:, 3] + correction_Sun_O_over_Fe) * unyt.dimensionless
+
+# Define the scatter as offset from the mean value
+y_scatter = unyt.unyt_array((Z_median - Z_lo, Z_hi - Z_median))
+
+# Meta-data
+comment = (
+    "Data obtained assuming a Chabrier IMF and h=0.7. "
+    f"h-corrected for SWIFT using cosmology: {cosmology.name}. "
+    "The metallicity is expressed as [alpha/Fe]. "
```
EvgeniiChaikin, line 59: Please indicate that alpha stands for Oxygen

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
```diff
@@ -0,0 +1,90 @@
...
+
+output_filename = "Gallazzi2021_Data.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Read the data
+raw = np.loadtxt(input_filename, delimiter=delimiter)
+M_star = 10 ** raw[:, 0] * unyt.Solar_Mass * (h_sim / h_obs) ** -2
```
EvgeniiChaikin, line 28: If I am reading it correctly,` Mstar `will have the units of **Msun h^-2** but you are telling `unyt` that the units of `Mstar` are just `Msun` (that is, without `h**2`)

Are you sure this is the correct way?

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
```diff
@@ -0,0 +1,90 @@
...
+redshift = 0.1
+h = h_sim
+
+# Write everything
+processed = ObservationalData()
+processed.associate_x(
+    M_star, scatter=None, comoving=True, description="Galaxy Stellar Mass"
+)
+processed.associate_y(
+    Z_median, scatter=y_scatter, comoving=True, description="Stellar metallicity"
```
EvgeniiChaikin, line 76: Stellar metallicity -> [alpha/Fe]*

EvgeniiChaikin (changes requested):
Hi @correac 

Thank you for adding this nice dataset!

I've had a look at your changes and added a few comments to your script.  There is one comment about the hubble parameter `h` - I am thinking that in the current version we may have two extra `h` in `Mstar` but I might also be wrong.

data/GalaxyStellarMassAlphatoIron/raw/gallazzi_2021_ascii.txt
```diff
@@ -0,0 +1,20 @@
+# mass-metallicity data from Gallazzi et al. 2021
```
EvgeniiChaikin, line 1: Could you please indicate which table or equation in the paper these data (numbers) come from?

EvgeniiChaikin:
In the current version, the Gallazzi+ (2021) data looks like this

stellar_mass_star_o_over_fe_100

Are the errors of the expected size?

data/GalaxyStellarMassAlphatoIron/raw/gallazzi_2021_ascii.txt
correac: These numbers come from bottom panel of Fig. 4.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Correct!

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Alpha does not stand for Oxygen. Gallazi et al. adopt a semi-empirical estimate of [𝛼/Fe] building on the work of Gallazzi et al. (2006). For each galaxy they measure the index ratio Mgb/⟨Fe⟩.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Yes, I want to directly convert from Gallazzi Msun convention directly to Msun.

Gallazzi gives stellar masses in units of Msun (h/0.7)^2. Now I think that the correct way to calculate M_star is : 
M_star = (
    10 ** raw[:, 0] * unyt.Solar_Mass * (h_obs / 0.7) ** 2
)

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Fixed.

correac:
The above figure shows the scatter calculated by Gallazzi, not error bars, but scatter. Since it is quite large, perhaps we can just show the median estimations.

data/GalaxyStellarMassAlphatoIron/raw/gallazzi_2021_ascii.txt
EvgeniiChaikin: Thanks!

EvgeniiChaikin:
> The above figure shows the scatter calculated by Gallazzi, not error bars, but scatter. Since it is quite large, perhaps we can just show the median estimations.

I don't understand why in this figure the scatter, e.g., at `Mstar = 1e10 Msun` **exceeds 0.4** along the Y-axis while the scatter in the bottom panel of Fig. 4 in Gallazzi's paper does not. Is this to be expected?

correac:
I think the figure shifts slightly in the x-axis due to the (h/h_obs)^2 factor, and shifts in the y-axis due to the conversion to Asplund+ solar metallicities.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
EvgeniiChaikin: Yes, it looks like multiplying by ` (h_obs / 0.7) ** 2` is the correct way.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
EvgeniiChaikin: Thanks, Camila!

Do you think we should also mention it in the comment (so that whoever uses these data in `swift-pipeline` knows what **alpha** is) or is it not necessary?

correac: Yes, better to add it in the comment. I will modify the branch adding the comment and fix the conversion to MStellar.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Fixed.

data/GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py
correac: Fixed

EvgeniiChaikin (approved):

bwvdnbro (approved):

PR #104, opened 2021-11-18
Replace 'means' with 'medians' in Saintonge2017 data

EvgeniiChaikin:
Replace 'means' with 'medians' in **Saintonge2017** data

- See table 5 on page 13 in https://arxiv.org/pdf/1710.02157.pdf and discussion on page 11

- We want Saintonge2017's medians (in the paper referred to as 'Binning' method), not means (in the paper referred to as 'Stacking method) because we use **median** when we plot data from our simulations (see, e.g., https://github.com/SWIFTSIM/pipeline-configs/blob/master/colibre/auto_plotter/gas_fraction_plots.yml#L214)

**BEFORE THE CHANGE**

h2_frac_func_stellar_mass

**AFTER THE CHANGE**
h2_frac_func_stellar_mass_new

### DIFF
--- a/data/GalaxyH2Fractions/conversion/convertSaintonge2017.py
+++ b/data/GalaxyH2Fractions/conversion/convertSaintonge2017.py
@@ -71,15 +71,16 @@ filetag = ["abcissa_M_star", "abcissa_mu_star", "abcissa_NUV_minus_r", "abcissa_
 for i in range(len(tables)):
     processed = ObservationalData()
 
+    # Take values from the 'Binning' method in Table 5 from Saintonge et al. (2017)
     if i != 2:
-        x_vals = 10 ** tables[i][:, 0] * units[i]
+        x_vals = 10 ** tables[i][:, 4] * units[i]
     else:
-        x_vals = tables[i][:, 0] * units[i]
+        x_vals = tables[i][:, 4] * units[i]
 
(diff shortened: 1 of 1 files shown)


EvgeniiChaikin:
@james-trayford could you please confirm that what I do in this PR is correct?

james-trayford:
@EvgeniiChaikin Thanks for revisiting this, a good find,  I agree the updated data is closer to what we're doing.

Now im thinking about this again I notice this is the lowest fH2 relation they provide, partially because this also includes upper limits. May think a bit about better comparing this, as upper limits of non detections may tend to pull down the observed relations from 'reality' because of the inclusion of fH2=0 galaxies (that would be removed from the median as 0s). This probably wouldn't be an issue for most bins, but at higher masses where there are more non-detections, can effect or set the median.

JBorrow:
What is the status of this?

EvgeniiChaikin:
If @james-trayford is happy with these changes, I am ready to merge them; according to his comment above, we may actually need more changes, though.

Since this PR is hanging for so long, maybe we could first merge these changes, and later - in a separate PR - add the other part with the corrections? @james-trayford what do you think?

james-trayford:
Hi @EvgeniiChaikin @JBorrow, apologies for the silence, I think it's best to merge this at this point as Evgenii proposes.

PR #110, opened 2022-02-07
Radio and IR SFRD data

bwvdnbro:
This adds SFRD data from Novak et al. (2017), Gruppioni et al. (2020) and Enia et al. (2022). The figure shows the new data (and Madau & Dickinson 2014 for reference):

SFRD

This can also be compared with the left panel in Fig. 10 of Enia et al. (2022).

### DIFF
--- a/data/StarFormationRateHistory/conversion/convertNovak2017.py
+++ b/data/StarFormationRateHistory/conversion/convertNovak2017.py
@@ -39,14 +39,22 @@ def cosmic_star_formation_history_novak():
     data = np.loadtxt(f"../raw/sfr_novak2017.dat")
 
     # Fetch the fields we need
-    z, z_minus, z_plus = data[:, 0], data[:, 1], data[:, 2]
-    SFR, SFR_minus, SFR_plus = data[:, 3], data[:, 4], data[:, 5]
+    z, delta_z_minus, delta_z_plus = data[:, 0], data[:, 1], data[:, 2]
+    SFR, delta_SFR_minus, delta_SFR_plus = data[:, 3], data[:, 4], data[:, 5]
 
-    z_bin = unyt.unyt_array(z, units="dimensionless")
-    z_scatter = unyt.unyt_array((-z_minus, z_plus), units="dimensionless")
+    a = 1.0 / (1.0 + z)
+    # division turns large into small, so minus <-> plus
+    a_minus = 1.0 / (1.0 + z + delta_z_plus)
+    a_plus = 1.0 / (1.0 + z + delta_z_minus)
+    delta_a_minus = a - a_minus
+    delta_a_plus = a_plus - a
+    print(delta_a_minus, delta_a_plus)
+
+    a_bin = unyt.unyt_array(a, units="dimensionless")
+    a_scatter = unyt.unyt_array((delta_a_minus, delta_a_plus), units="dimensionless")
     # convert from log10(SFRD) to SFRD and carry the uncertainties
-    SFR_minus = 10.0 ** (SFR + SFR_minus)
-    SFR_plus = 10.0 ** (SFR + SFR_plus)
+    SFR_minus = 10.0 ** (SFR + delta_SFR_minus)
+    SFR_plus = 10.0 ** (SFR + delta_SFR_plus)
     SFR = 10.0 ** SFR
     SFR_scatter = unyt.unyt_array(
         (SFR - SFR_minus, SFR_plus - SFR), units="Msun/yr/Mpc**3"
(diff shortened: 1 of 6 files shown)


EvgeniiChaikin:
Thanks, @bwvdnbro, the new data look great! I will do a full review later today.

data/StarFormationRateHistory/conversion/convertNovak2017.py
```diff
@@ -0,0 +1,87 @@
...
+    processed.associate_cosmology(cosmology)
+
+    if not os.path.exists(output_directory):
+        os.mkdir(output_directory)
+
+    # Load raw Enia2022 data
+    data = np.loadtxt(f"../raw/sfr_novak2017.dat")
+
+    # Fetch the fields we need
+    z, z_minus, z_plus = data[:, 0], data[:, 1], data[:, 2]
```
EvgeniiChaikin, line 42: Unlike in the other two datasets, I guess here you are reading `delta_z_minus` and `delta_z_plus` rather than `z_minus `and `z_plus`? (because later you don't subtract `z` from them, unlike in the other two cases)

Perhaps, adding a prefix 'delta' here to the names of the two variables will slightly improve the readability of the code, if you agree

EvgeniiChaikin (commented):
I don't have much to say about the code; it all looks good and easy to read. I left only one comment about notation in one of the 3 scrips.

_A more general note:_
In the current cosmic SFH script, we use the scale factor `a` as an independent variable, not redshift `z` (even though what we show along the x-axis is redshift).

So when I was testing this PR, in the SFH script I had to write code like this

```
Enia2022 = load_observations(
    f"{path_to_obs_data}/data/StarFormationRateHistory/Enia2022.hdf5"
)[0]
Gruppioni2020 = load_observations(
    f"{path_to_obs_data}/data/StarFormationRateHistory/Gruppioni2020.hdf5"
)[0]
Novak2017 = load_observations(
    f"{path_to_obs_data}/data/StarFormationRateHistory/Novak2017.hdf5"
)[0]

for data, color in zip(
    [Enia2022, Gruppioni2020, Novak2017], ["maroon", "goldenrod", "slategrey"]
):

    # data's x is redshift, but plot's x is scale factor
    a = 1.0 / (1.0 + data.x.value)
    a_plus = 1.0 / (1.0 + data.x.value - data.x_scatter[0].value)
    a_minus = 1.0 / (1.0 + data.x.value + data.x_scatter[1].value)

    # both deltas are positive
    delta_a_plus = a_plus - a
    delta_a_minus = a - a_minus

    SFR = data.y.value
    observation_lines.append(
        plt.errorbar(
            a,
            SFR,
            yerr=[data.y_scatter[0].value, data.y_scatter[1].value],
            xerr=[delta_a_minus, delta_a_plus],
            label=data.citation,
            linestyle="None",
            zorder=-100,
            color=color,
        )
    )
    observation_labels.append(data.citation)
```

Might be a bit too much unnecessary code for the plotting script, but I was able to reproduce your plot

star_formation_history

data/StarFormationRateHistory/conversion/convertNovak2017.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+    processed.associate_y(
+        SFR,
+        scatter=SFR_scatter,
+        comoving=False,
+        description="Cosmic average star formation rate density",
+    )
+
+    z_minus = (z + z_minus).min()
+    z_plus = (z + z_plus).max()
```
EvgeniiChaikin, line 70: regarding my other comment, note that here the LHS is the true `z_plus`, but `z_plus` on the right is `delta_z_plus` 

and the same logic applies to the line above with `z_minus`

data/StarFormationRateHistory/conversion/convertNovak2017.py
bwvdnbro: Absolutely! It is always better to avoid confusion. Done.

data/StarFormationRateHistory/conversion/convertNovak2017.py
```diff
@@ -0,0 +1,95 @@
...
+    z, delta_z_minus, delta_z_plus = data[:, 0], data[:, 1], data[:, 2]
+    SFR, delta_SFR_minus, delta_SFR_plus = data[:, 3], data[:, 4], data[:, 5]
+
+    a = 1.0 / (1.0 + z)
+    # division turns large into small, so minus <-> plus
+    a_minus = 1.0 / (1.0 + z + delta_z_plus)
+    a_plus = 1.0 / (1.0 + z + delta_z_minus)
+    delta_a_minus = a - a_minus
+    delta_a_plus = a_plus - a
+    print(delta_a_minus, delta_a_plus)
```
EvgeniiChaikin, line 51:
```suggestion
```

EvgeniiChaikin:
> I have converted redshift to scale factor to make inclusion in the SFH script easier.

Thank you @bwvdnbro! Everything looks perfect! I only added one last suggestion for removing a debug-related line.

PR #111, opened 2022-02-11
Wrote parallel conversion script

bwvdnbro:
I got a bit tired of waiting 8 minutes for the conversion script to finish on my desktop. Running the individual conversion scripts can easily be done in parallel.
I have converted the bash script into a Python script that uses `subprocess` to run `NPROC` scripts in parallel, with a new optional argument `--nproc` controlling the number of parallel processes (default: 1).

This uses the same technique used in the new task plot scripts in the SWIFT repository (see https://github.com/SWIFTSIM/swiftsim/blob/master/tools/task_plots/process_plot_tasks.py).

The runtime on my desktop went from 8 minutes to 2 minutes with `--nproc 4` and 1 minute with `--nproc 16`. So pretty reasonable scaling (my desktop has 12 cores).


bwvdnbro:
Using 2 processes for the CI (CI workers should have 2 cores) yields a disappointingly small speedup of about 15%.

bwvdnbro:
In its current form, we could probably set up some complicated GNU parallel command that does the same. The Python version has the advantage that you can easily add other commands to it or sort commands based on some weight for improved performance (like is done for the task plots in SWIFT). And I think it is easier to understand the script.

Also: GNU parallel is not necessarily available on all systems. Python would need to be anyway, so it might be cleaner this way.

PR #115, opened 2022-06-02
FIREbox data

bwvdnbro:
This PR adds reference data from the FIREbox simulation (Feldmann et al., 2022).

Included are:
 - Mstar vs SFR (Fig. 5)
 - Mstar vs passive fraction (Fig. 6)
 - Mstar vs MHI/MH2 (Fig. 7)
 - Mstar vs [O/H] for gas (Fig. 8)
 - Mstar vs [Fe/H] for stars (Fig. 8)
 - Stellar mass function (Fig. 9)
 - Mstar vs Mstar/Mhalo (Fig. 10)
 - Cosmic star formation density history (Fig. 11)
 - Cosmic stellar density history (Fig. 11)
 - Cosmic HI and H2 evolution (Fig. 12)
 - HI Column Density Distribution Function (Fig. 14)

All data were extracted using WebPlotDigitizer and the images as provided in the arXiv source. Apart from the CDDF data, which was pretty much extracted by eye, all other data should be reasonably reliable.


data/CosmicHIAbundance/conversion/convertFIREbox.py
```diff
@@ -0,0 +1,52 @@
...
+citation = "Feldmann et al., 2022 (FIREbox)"
+bibcode = "2022arXiv220515325F"
+name = "HI History"
+plot_as = "line"
+
+a = raw[:, 0]
+z = 1.0 / a - 1.0
+rhoHI = 10.0 ** raw[:, 1]
+a = unyt.unyt_array(a, units=unyt.dimensionless)
+rhoHI = unyt.unyt_array(rhoHI, units=unyt.Msun / unyt.Mpc ** 3)
```
EvgeniiChaikin, line 27: Shouldn't the units be Msun / **comoving** Mpc**3?

data/ColumnDensityDistributionFunction/conversion/convertFIREbox.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import sys
+
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+# Meta-data
+name = f"CDDF from Ho et al. (2021)"
```
EvgeniiChaikin, line 13: Is the name correct?

EvgeniiChaikin (commented):
Thank you very much for this upgrade @bwvdnbro !

I was able to plot all data from this PR without any changes in the code and everything looked as expected.

I have left a couple of minor comments in the code, most of which are simply to improve the description of the data.

I also have a few more general comments:

- For the observational data that is available at **multiple** redshifts, we should save it as a single file. This will enable velociraptor to choose the data with the closest redshift to the snapshot. You can see how this is done e.g. here (https://github.com/SWIFTSIM/velociraptor-comparison-data/blob/master/data/GalaxyStellarMassPassiveFraction/conversion/convertBehroozi2019.py)

- Do we want to plot the FIREbox data as points or as lines? (currently it is shown as points).

- I don't think we need redshift suffixes in the names of the generated .hdf5 files with the FIREbox data.

- In the citation fields, I would write `Feldmann et al. (2022, FIREbox)' instead of `Feldmann et al., 2022 (FIREbox)'

data/CosmicHIAbundance/conversion/convertFIREbox.py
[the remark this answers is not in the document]
bwvdnbro: Yes, but I'm not sure how to tell `unyt` about that. The plotting script in the pipeline that uses these data plots the HI mass from the statistics file divided by the comoving box size, which is the same quantity. These units are not actually used anywhere, I think.

data/GalaxyH2Fractions/conversion/convertFIREbox.py
```diff
@@ -0,0 +1,51 @@
...
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+input_filename = f"../raw/FIREbox_z0.txt"
+
+processed = ObservationalData()
+raw = np.loadtxt(input_filename)
+
+comment = "Showing all central galaxies, using a 10 kpc aperture."
```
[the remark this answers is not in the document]
bwvdnbro: I have to say that I don't like this. M_H2 is the mass of molecular hydrogen, which per definition does not include any Helium. So I would say that this plot is mislabelled and we are actually showing a different quantity than what the y label axis claims (we had a similar discussion for https://github.com/SWIFTSIM/pipeline-configs/pull/163).

Anyway, you are right that we should compare apples with apples, so I will multiply the FIREbox data with 1.36 and add some more information to the description. Whether or not those data should then be called M_H2 is not really relevant for this PR.

data/ColumnDensityDistributionFunction/conversion/convertFIREbox.py
bwvdnbro: No, that is a bad case of copy-paste.

EvgeniiChaikin:
@bwvdnbro Thank you for the update! I think it all looks good now! I don't have any further comments.

Some updated plots are below

column_density_distribution_function_chunk1_reduced_range
hi_frac_func_stellar_mass
HI_mass_evolution
stellar_mass_function_30
stellar_mass_gas_sf_metallicity_lom_allgals_30
stellar_mass_specific_sfr_active_50
stellar_mass_with_scatter_passive_fraction_50
h2_frac_func_stellar_mass
stellar_mass_specific_sfr_all_50
stellar_mass_star_fe_over_h_mol_lofloor_50

bwvdnbro:
Okay, thanks for your review!

Since you already have these plots, I assume you have some version of the pipeline that already uses these data? Would you be able to open a PR for the new data in the pipeline repository?

PR #120, opened 2022-07-07
Draft: Cold gas evolution data conversion

bwvdnbro:
Since I needed an easy way to add the cosmic HI/H2 abundance data to the new hypercube plots, I started implementing proper conversion scripts for these data, rather than reading the `.txt.` files directly from the plotting pipeline. This has the added advantage that we no longer need to obtain the cosmology from the snapshot, but can just use the `cosmology` that is passed on to the conversion script.

### DIFF
--- a/data/CosmicHIAbundance/conversion/convertPeroux2020.py
+++ b/data/CosmicHIAbundance/conversion/convertPeroux2020.py
@@ -1,0 +1,75 @@
+from velociraptor.observations.objects import ObservationalData
+
+import unyt
+import numpy as np
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+input_filename = "../raw/Peroux2020_OmegaHI.txt"
+
+processed = ObservationalData()
+
+comment = "Based on a fit."
+citation = "Peroux & Howk (2020)"
+bibcode = "2020ARA&A..58..363P"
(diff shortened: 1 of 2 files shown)


MatthieuSchaller:
We should document the cosmology correction probably in the help string. That's something we added.

data/CosmicHIAbundance/conversion/convertPeroux2020.py
```diff
@@ -0,0 +1,75 @@
...
+rhoHI = fitting_formula(zgrid, *raw[0])
+rhoHI_low = fitting_formula(zgrid, *raw[1])
+rhoHI_high = fitting_formula(zgrid, *raw[2])
+
+cosmology_correction = (
+    np.sqrt(cosmology.Om0 * (1.0 + zgrid) ** 3 + cosmology.Ode0) / cosmology.h
+) / (np.sqrt(0.3 * (1.0 + zgrid) ** 3 + 0.7) / 0.7)
+rhoHI *= 0.76 * cosmology_correction
+rhoHI_low *= 0.76 * cosmology_correction
+rhoHI_high *= 0.76 * cosmology_correction
```
MatthieuSchaller, line 44: Same with this Helium correction.

bwvdnbro:
Done. Anything else you want changed?

MatthieuSchaller (approved):
That's now good for me. 

I'll let you merge it however if it is still relevant. I realise I should have acted earlier.

PR #125, opened 2022-08-01
Include some spatially-resolved relations of Abdurro'uf + 2022

Fonotec:
@bwvdnbro these are the Kennicutt-Schmidt relations, but also the molecular gas main sequence and the star formation main sequence binned for the new paper of Abdurro'uf + (2022). 

Let me know if you have suggestions.


data/SpatiallyResolvedMolecularGasMainSequence/convert/convertAbdurrouf2022.py
```diff
@@ -0,0 +1,104 @@
...
+
+array_of_interest = np.arange(-1, 4, 0.25)
+minimum_surface_density = 1.0  # in their paper this seems to be the limit they can observe, however they do not say anything about this
+array_of_interest = array_of_interest[array_of_interest >= minimum_surface_density]
+if array_of_interest[0] > minimum_surface_density:
+    array_of_interest = np.append([minimum_surface_density], array_of_interest)
+
+Sigma_star, Sigma_H2 = np.genfromtxt(input_filename, unpack=True)
+
+Sigma_H2 = 10 ** (Sigma_H2-6.0)
```
bwvdnbro, line 53: I'm not a big fan of this manual unit conversion. The whole point of using a `unyt_array` with units below is that you don't need to do these conversions manually. The rest of the pipeline knows how to deal with these units. So if the data is in `Mpc/kpc^2`, just use that as a unit.

data/SpatiallyResolvedNeutralKSRelation/conversion/convertAbdurrouf2022.py
```diff
@@ -0,0 +1,104 @@
...
+minimum_surface_density = 0.5  # in their paper this seems to be the limit they can observe, however they do not say anything about this
+array_of_interest = array_of_interest[array_of_interest >= minimum_surface_density]
+if array_of_interest[0] > minimum_surface_density:
+    array_of_interest = np.append([minimum_surface_density], array_of_interest)
+
+Sigma_Hn, Sigma_SFR = np.genfromtxt(input_filename, unpack=True)
+
+Sigma_SFR = 10 ** (Sigma_SFR)
+
+Sigma_Hn = (10 ** (Sigma_Hn - 6.0)) / 1.36
```
bwvdnbro, line 55: Here it would also be useful to add a little comment about the 1.36 correction factor.

bwvdnbro (changes requested):
Looks good! I have left some small comments (the first one applies to all the scripts).

data/SpatiallyResolvedNeutralKSRelation/conversion/convertAbdurrouf2022.py
Fonotec: I left a comment at that line.

data/SpatiallyResolvedMolecularGasMainSequence/convert/convertAbdurrouf2022.py
Fonotec: Okay, I did that

Fonotec:
I forgot to put the bin arrays in the right units.

Fonotec:
Now it is good to go

bwvdnbro (approved):

PR #128, opened 2022-08-11
Spatially-resolved Pessa+(2021) data

Fonotec:
These are the molecular gas main sequence and the star formation main sequence binned for the Pessa+(2021) paper. 

@bwvdnbro let me know what you think.


bwvdnbro (commented):
This looks good to me. But do we really need to have the 100pc and 1000pc files in the repository if we are only using the 500pc data?

Fonotec:
For now maybe not, but if we look at smaller regions I think it is worth having the option to use 100 pc (or larger use the 1kpc). We want to compare the simulations with the closest binning size in the observation. So if we want to look at high surface densities, we have to use smaller binning sizes because otherwise peaks are too much smoothed/averaged out and we want a certain number of bins at high surface densities. 

Besides this, this data set is unique because it is the only one I am aware of that shows the relations at different binning sizes. To add, if it is in the repository everyone has access to it. Currently, it is just in an email. 

Ideally, it would maybe be good to have the relation at different binning sizes but without a lot of duplicate code.

bwvdnbro:
In that case you could convert all of them using the one script and different names. That way, we can easily change the data in the other pipeline by just reading a different file. That saves you from having the make another change to this repository if we ever decide to change the data in the pipeline.

Fonotec:
Okay, that sounds good to me, I will add this change. I will do the same for the Pessa+(2022) molecular KS relation because now it also only reads the 500 pc version.

Fonotec:
Bert, this includes the suggestion of converting 3 separate files for the 3 resolutions.

bwvdnbro (approved):

PR #130, opened 2022-08-18
Extra BH plots

FHusko:
Add observational data for BH mass function, BH spin - mass relation and AGN radio luminosity function


data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
```diff
@@ -0,0 +1,69 @@
+from velociraptor.observations.objects import ObservationalData
+from astropy.cosmology import WMAP7 as cosmology
+
+import unyt
+import numpy as np
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
```
bwvdnbro, line 9: You are really not supposed to comment out these lines and import your own cosmology.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
```diff
@@ -0,0 +1,69 @@
...
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+processed = ObservationalData()
+
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10**raw[:, 0] * unyt.dimensionless
+L1_4_low = 10**(raw[:, 0] -raw[:, 2]) * unyt.dimensionless
```
bwvdnbro, line 30: In the raw data file, these quantities do have units, so `unyt.dimensionless` is incorrect.

Note that it is important to have these units in place, because those are used by the AutoPlotter plots in the pipeline (we will get to that in my review for your PR in the other repository).

bwvdnbro (changes requested):
This generally looks okay, except for two general and important comments:

 - you should uncomment the three lines that import the cosmology and not import your own astropy.cosmology. The cosmology script is in the main repository folder, so to run the scripts, you simply use `python SCRIPT_NAME ../../../cosmology.py`. That is also what the conversion script does.
 - We need the correct units. Now you used `unyt.dimensionless` everywhere, which is incorrect. With the correct units in place, the converted files can be directly imported by the AutoPlotter, which makes it a lot easier to include them in pipeline plots.

Once you made these changes, I will go through the scripts again to see if there is anything else that needs changing.

bwvdnbro:
Also: as you can see from the failed CI, you need to run the formatting script.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
FHusko: This was left over from when I was creating the inidividual raw data files and conversion scripts and testing them out one by one - I think I had issues there since I wasn't passing any cosmology to the conversion scripts.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
FHusko: The units should now be correct, and the scripts all seem to run okay.

FHusko:
> Also: as you can see from the failed CI, you need to run the formatting script.

Formatting script has now been run.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
```diff
@@ -0,0 +1,75 @@
...
+processed = ObservationalData()
+
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.Watt / unyt.Hertz
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.Watt / unyt.Hertz
+L1_4_high = 10 ** (raw[:, 0] + raw[:, 1]) * unyt.Watt / unyt.Hertz
+
+Phi = 10 ** raw[:, 3] * unyt.Mpc ** -3
```
bwvdnbro, line 32: This line made me cringe a bit, although it is apparently okay to do this in Python.

Still, could you either put brackets around the `-3` exponent or use a division (`/unyt.Mpc**3`) instead? `** -` does not work in most programming languages, so I'm not a big fan.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
```diff
@@ -0,0 +1,75 @@
...
+comment = (
+    " AGN radio luminosity data, taken from Best & Heckmann (2012):"
+    " 2012MNRAS.421.1569B"
+)
+citation = "Best & Heckmann (2012)"
+bibcode = "2012MNRAS.421.1569B"
+name = "AGN Radio Luminosity Function"
+plot_as = "points"
+redshift = 0.1
+h = cosmology.h
```
bwvdnbro, line 49: Why is `h` defined? It does not look like it is used anywhere...

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+input_filename = "../raw/BestHeckmann2012.txt"
+delimiter = None
+half_mass = 1
```
bwvdnbro, line 14: It looks like `half_mass` and `log_mass` are never used, so why are they defined here?

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSadler2002.py
```diff
@@ -0,0 +1,75 @@
...
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+processed = ObservationalData()
+
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.dimensionless
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.dimensionless
```
bwvdnbro, line 29: Are these dimensionless?

According to the data text file:
```
# Units of the luminosity are W Hz^-1. Units of the luminosity 
# function are W Hz^-1 Mpc^-3 dex^-1
```

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
```diff
@@ -0,0 +1,77 @@
+from velociraptor.observations.objects import ObservationalData
+from astropy.cosmology import WMAP7 as cosmology
```
bwvdnbro, line 2: This line is no longer needed, or at least should not be.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2017.py
```diff
@@ -0,0 +1,75 @@
...
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+processed = ObservationalData()
+
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.dimensionless
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.dimensionless
```
bwvdnbro, line 29: I don't think these are dimensionless either.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
```diff
@@ -0,0 +1,77 @@
...
+
+comment = (
+    " AGN radio luminosity data, taken from Smolcic et al. (2009):"
+    " 2009ApJ...696...24S"
+)
+citation = "Smolcic et al. (2009)"
+bibcode = "2009ApJ...696...24S"
+name = "AGN Radio Luminosity Function"
+plot_as = "points"
+redshift = 0.23
```
bwvdnbro, line 50: The redshift is used to determine when observational data should be included when we use the Autoplotter in the pipeline. It is therefore better to allow for a small scatter around a target redshift, especially if you have a very precise value like this (since it is quite unlikely that we will have a snapshot at exactly this redshift).

Most conversion scripts use a variant of `associate_redshift()` that takes three arguments: the central redshift and the upper and lower limit of a redshift interval. I am not saying you have to do that here, but I wanted to make you aware of this.

bwvdnbro (changes requested):
This is almost ready. A few more points:
 - all your scripts define variables `log_mass`, `half_mass` and `h` that are never used. Please remove those
 - I think there are still some data that are labelled as dimensionless but actually have dimensions. Can you check these again?

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
FHusko: Indeed my instinct is not to do this, but Python has given me this bad habit... I have replaced instances of `*unyt.Mpc**-3` with `/unyt.Mpc**3` now.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
FHusko: I think this may have been carried over from whatever script I was using as a template. I have removed this line now in all scripts but one, where I introduce it a bit earlier, because I actually need to use `h` to convert the observational data.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertBestHeckmann2012.py
FHusko: Same story as `h` above, or in this case I may have introduced these to use them somewhere but then realized I didn't need them, or something along those lines (I don't remember introducing them at all).

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSadler2002.py
FHusko: Indeed these should all be `unyt.Watt / unyt.Hertz` now, thanks for catching that.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
FHusko: Yes, thanks for catching it.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2017.py
FHusko: Indeed, only the spins in the black hole spin - mass relation should now be dimensionless.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
FHusko: Thanks for letting me know about this. This is useful, especially for the AGN radio luminosity functions, which are typically far from z = 0 or z = 0.1 (more like z = 0.1 - 0.4).

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertMauchSadler2007.py
```diff
@@ -0,0 +1,74 @@
...
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.Watt / unyt.Hertz
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.Watt / unyt.Hertz
+L1_4_high = 10 ** (raw[:, 0] + raw[:, 1]) * unyt.Watt / unyt.Hertz
+
+Phi = 2.51 * 10 ** raw[:, 3] / unyt.Mpc ** 3
+Phi_low = 2.51 * 10 ** (raw[:, 3] - raw[:, 5]) / unyt.Mpc ** 3
+Phi_high = 2.51 * 10 ** (raw[:, 3] + raw[:, 4]) / unyt.Mpc ** 3
```
bwvdnbro, line 32: I seem to have glossed over this before, but what is the factor `2.51` here? Could you add a comment explaining where this factor comes from?

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertPadovani2015.py
```diff
@@ -0,0 +1,74 @@
...
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.Watt / unyt.Hertz
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.Watt / unyt.Hertz
+L1_4_high = 10 ** (raw[:, 0] + raw[:, 1]) * unyt.Watt / unyt.Hertz
+
+Phi = 2.31 * 10 ** -9 * 10 ** raw[:, 3] / unyt.Mpc ** 3
+Phi_low = 2.31 * 10 ** -9 * 10 ** (raw[:, 3] - raw[:, 5]) / unyt.Mpc ** 3
+Phi_high = 2.31 * 10 ** -9 * 10 ** (raw[:, 3] + raw[:, 4]) / unyt.Mpc ** 3
```
bwvdnbro, line 32: Same question about the factor `2.31 * 10 ** -9` here. Also, since this is probably one factor, you might want to write `2.31e-9`, which is more compact and therefore clearer.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
```diff
@@ -0,0 +1,74 @@
...
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = raw[:, 0] * (float(1e21) * unyt.Watt / unyt.Hertz)
+L1_4_low = (raw[:, 0] - raw[:, 2]) * (float(1e21) * unyt.Watt / unyt.Hertz)
+L1_4_high = (raw[:, 0] + raw[:, 1]) * (float(1e21) * unyt.Watt / unyt.Hertz)
+
+Phi = raw[:, 3] * (float(1e-4) / unyt.Mpc ** 3)
+Phi_low = (raw[:, 3] - raw[:, 5]) * (float(1e-4) / unyt.Mpc ** 3)
+Phi_high = (raw[:, 3] + raw[:, 4]) * (float(1e-4) / unyt.Mpc ** 3)
```
bwvdnbro, line 32: I don't think the `float()` is necessary here, you can just write `1.e-4`.

data/BlackHoleMassFunction/conversion/convertMarconi2004.py
```diff
@@ -0,0 +1,78 @@
...
+
+Phi = 10 ** raw[:, 3] / unyt.Mpc ** 3
+Phi_low = 10 ** (raw[:, 5]) / unyt.Mpc ** 3
+Phi_high = 10 ** (raw[:, 4]) / unyt.Mpc ** 3
+
+# Define the scatter as offset from the mean value
+x_scatter = unyt.unyt_array((M_BH - M_BH_low, M_BH_high - M_BH))
+y_scatter = unyt.unyt_array(
+    (
+        -(10 ** (2 * np.log10(Phi) - np.log10(Phi_high))) / unyt.Mpc ** 3 + Phi,
```
bwvdnbro, line 38: Why this weird expression? Should this not just be `Phi - Phi_low`?

data/BlackHoleMassFunction/conversion/convertMerloniHeinz2008.py
```diff
@@ -0,0 +1,75 @@
...
+
+Phi = 10 ** raw[:, 3] / unyt.Mpc ** 3
+Phi_low = 10 ** (raw[:, 5]) / unyt.Mpc ** 3
+Phi_high = 10 ** (raw[:, 4]) / unyt.Mpc ** 3
+
+# Define the scatter as offset from the mean value
+x_scatter = unyt.unyt_array((M_BH - M_BH_low, M_BH_high - M_BH))
+y_scatter = unyt.unyt_array(
+    (
+        -(10 ** (2 * np.log10(Phi) - np.log10(Phi_high))) / unyt.Mpc ** 3 + Phi,
```
bwvdnbro, line 38: Same here.

bwvdnbro (changes requested):
There are a few other things I just noticed that also need some clarification. I think this should be the last iteration before this can be merged.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertMauchSadler2007.py
FHusko: I have added comments explaining this and other factors. This one is a conversion from log space in base 2.5 (magnitudes) to decades in base space 10. I.e. it is log_base_2.5(10). I also now explicitly calculate it in the script, instead of manually inserting the value. Same goes for similar factors that appear in other scripts.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertPadovani2015.py
FHusko: This one is a conversion from the natural logarithm to base 10, i.e. it is ln(10). The `10 ** - 9` is a conversion factor from Gpc^-3 to Mpc^-3.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertSmolcic2009.py
FHusko: I have removed the `float()`s.

data/BlackHoleMassFunction/conversion/convertMarconi2004.py
FHusko: In principle, yes. However, when extracting the observational data, I decided to (out of simplicity and to save time) use only the high error bars, and then calculate the low ones from the high ones, since the errors are symmetrical in the papers in question. One complication is that the data is given in linear space, but the errors are symmetric in log space in the papers.

I have now replaced the previous weird expression with
`y_scatter_low = Phi - Phi * 10 ** (np.log10(Phi) - np.log10(Phi_high))`
which perhaps looks a bit more intuitive (but is the same). Essentially what I am doing is calculating the log-space error from the mean and the high point as `-np.log10(Phi) + np.log10(Phi_high)`, and then using that to define the low point so that it is symmetric on a log plot.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertPadovani2015.py
```diff
@@ -0,0 +1,82 @@
...
+# Read the data (only those columns we need here)
+raw = np.loadtxt(input_filename, delimiter=delimiter, usecols=(0, 1, 2, 3, 4, 5))
+
+L1_4 = 10 ** raw[:, 0] * unyt.Watt / unyt.Hertz
+L1_4_low = 10 ** (raw[:, 0] - raw[:, 2]) * unyt.Watt / unyt.Hertz
+L1_4_high = 10 ** (raw[:, 0] + raw[:, 1]) * unyt.Watt / unyt.Hertz
+
+# Padovani et al. (2015) define their luminosity function using the natural logarithm.
+# We convert to the usual definition, using dex, by multiplying with a conversion factor
+# that is equal to ln(10). We also multiply by 1e-9 to convert from units of Gpc^-3 to Mpc^-3.
```
bwvdnbro, line 32: Why not just use `unyt.Gpc**3` as units then? The whole point of using `unyt` is that it saves you from having to do manual unit conversions...

data/BlackHoleMassFunction/conversion/convertMarconi2004.py
bwvdnbro: Right, I think I understand now. I also noticed now that the lower error bounds are actually zero in the raw files.

In that case, it is better if you could not read that column from the raw files and also rename the error variable to `Phi_log_err` or something. Then it is clear that this error is to be used as a symmetric logarithmic error. You can then compute `Phi_log_err` from the data you read and compute the lower and upper bounds in a symmetric way, so that it looks cleaner.

FHusko: I have introduced this now for cases with a symmetric error.

data/BlackHoleAGNRadioLuminosityFunction/conversion/convertPadovani2015.py
FHusko: True enough, I have replaced the `unyt.Mpc**3` with `unyt.Gpc**3` and removed the numerical conversion factor.

bwvdnbro (approved):

FHusko:
> I think this is good to go now (provided that the CI passes). Unless there is anything else you wanted to change?

Not really, I am happy with this version. :)

PR #132, opened 2022-08-26
Added Strom+2022 gas metallicity data.

bwvdnbro:
Adds the 3 fits from Strom et al. (2022) for high redshift gas metallicity (https://ui.adsabs.harvard.edu/abs/2022ApJ...925..116S/abstract).

### DIFF
--- a/data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
+++ b/data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
@@ -1,0 +1,80 @@
+from velociraptor.observations.objects import ObservationalData
+
+import unyt
+import numpy as np
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Meta-data
+comment = "The median stellar mass is 1e10 Msun."
+citation = "Strom et al. (2022)"
+bibcode = "2022ApJ...925..116S"
+plot_as = "line"
+redshift = np.array([2.0, 2.7])
+
+M_star = unyt.unyt_array(np.logspace(9.0, 11.3, 100), units="Msun")
+
+models = {
+    "OH": {"alpha": (8.35, 0.02), "beta": (0.14, 0.05), "label": "12+log10(O/H)"},
+    "NH": {"alpha": (7.07, 0.03), "beta": (0.29, 0.07), "label": "12+log10(N/H)"},
+    "FeH": {"alpha": (-0.69, 0.02), "beta": (0.17, 0.22), "label": "[Fe/H]"},
+}
+
+
+def fitcurve(M, alpha, beta):
+    return alpha + beta * (np.log10(M / unyt.Msun) - 10.0)
+
+
+for model in models:
+    label = models[model]["label"]
+
+    name = f"Stellar mass - Gas phase metallicity relation ({label})"
+    output_filename = f"Strom2022_{model}.hdf5"
+
+    alpha, alpha_err = models[model]["alpha"]
+    beta, beta_err = models[model]["beta"]
+
+    Z = fitcurve(M_star, alpha, beta)
+    Zpp = fitcurve(M_star, alpha + alpha_err, beta + beta_err)
+    Zpm = fitcurve(M_star, alpha + alpha_err, beta - beta_err)
+    Zmp = fitcurve(M_star, alpha - alpha_err, beta + beta_err)
+    Zmm = fitcurve(M_star, alpha - alpha_err, beta - beta_err)
+    Z_max = np.maximum(np.maximum(Zpp, Zpm), np.maximum(Zmp, Zmm))
+    Z_min = np.minimum(np.minimum(Zpp, Zpm), np.minimum(Zmp, Zmm))
+
+    Z_scatter = unyt.unyt_array([Z - Z_min, Z_max - Z], units="dimensionless")
+    Z = unyt.unyt_array(Z, units="dimensionless")
+
+    # Write everything
+    processed = ObservationalData()
+    processed.associate_x(
+        M_star, scatter=None, comoving=False, description="Galaxy Stellar Mass"
+    )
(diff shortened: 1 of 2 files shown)


data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+    os.mkdir(output_directory)
+
+# Meta-data
+comment = "The median stellar mass is 1e10 Msun."
+citation = "Strom et al. (2022)"
+bibcode = "2022ApJ...925..116S"
+plot_as = "line"
+redshift = np.array([2.0, 2.7])
+
+M_star = unyt.unyt_array(np.logspace(9.0, 11.3, 100), units="Msun")
```
EvgeniiChaikin, line 24:
```suggestion
M_star = unyt.unyt_array(np.logspace(9.0, 11.3, 20), units="Msun")
```
With so many points as 100 it is impossible to distinguish individual error bars. 

stellar_mass_star_fe_over_h_mol_lofloor_50_old

Le'ts make it 20. Then the error bars will look as follows

stellar_mass_star_fe_over_h_mol_lofloor_50

data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+
+models = {
+    "OH": {"alpha": (8.35, 0.02), "beta": (0.14, 0.05), "label": "12+log10(O/H)"},
+    "NH": {"alpha": (7.07, 0.03), "beta": (0.29, 0.07), "label": "12+log10(N/H)"},
+    "FeH": {"alpha": (-0.69, 0.02), "beta": (0.17, 0.22), "label": "[Fe/H]"},
```
EvgeniiChaikin, line 29: It should be
```suggestion
    "FeH": {"alpha": (-0.69, 0.02), "beta": (0.17, 0.05), "label": "[Fe/H]"},
```
according to Table 1 from https://arxiv.org/pdf/2111.06410.pdf

data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Meta-data
+comment = "The median stellar mass is 1e10 Msun."
```
EvgeniiChaikin, line 18: I would also indicate in the comment that 

- This is **a fit** to the data containing 195 star-forming high-redshift galaxies
- Strom et al (2022) use solar abundancies from Asplund et al. (2009), so no corrections are needed (because for Colibre we also use Asplund et al. (2009))

data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+citation = "Strom et al. (2022)"
+bibcode = "2022ApJ...925..116S"
+plot_as = "line"
+redshift = np.array([2.0, 2.7])
```
EvgeniiChaikin, line 22: Let's expand the range a little bit, so whenever we make plots at z \approx 2 or z \approx 3, these data are always included.
```suggestion
redshift = np.array([1.9, 3.1])
```

data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def fitcurve(M, alpha, beta):
+    return alpha + beta * (np.log10(M / unyt.Msun) - 10.0)
+
+
+for model in models:
+    label = models[model]["label"]
+
+    name = f"Stellar mass - Gas phase metallicity relation ({label})"
```
EvgeniiChaikin, line 40: **OH** and **NH** are indeed gas-phase metallicities but **FeH** is stellar metallicity.

```suggestion
    name = f"Stellar mass - Metallicity relation ({label})"
```

data/GalaxyStellarMassGasMetallicity/conversion/convertStrom2022.py
```diff
@@ -0,0 +1,80 @@
...
+    # Write everything
+    processed = ObservationalData()
+    processed.associate_x(
+        M_star, scatter=None, comoving=False, description="Galaxy Stellar Mass"
+    )
+    processed.associate_y(
+        Z,
+        scatter=Z_scatter,
+        comoving=False,
+        description=f"Gas phase metallicity ({label})",
```
EvgeniiChaikin, line 66: See my comment above
```suggestion
        description=f"Metallicity ({label})",
```

EvgeniiChaikin (changes requested):
Thank you for the update @bwvdnbro 

Some small comments are below in the code and I also have two more important comments:

1) This script is currently in the directory in /GalaxyStellarMass**Gas**Metallicity, which is not quite right because in addition to gas-phase metallicities (OH and NH) it also converts data for stellar metallicity (FeH). I think the easiest way to resolve it is to create a second version of this script, put the second version in /GalaxyStellarMass**Stellar**Metallicity and make it only convert FeH; while in this version convert only OH and NH.

2) The gas-phase metallicity in our pipeline is shown as 12 + log(O/H) so the data produced by this script can be directly shown in those plots. However, the plots with stellar metallicity in the pipeline show Z/Z_sun (or in units of [Fe/H], 10 ** [Fe/H]), whereas this script gives [Fe/H] (or log Z/Z_sun). This means that in order to put Strom's FeH data on our plots with stellar metallicity, we will need to transform the y values produced by this script as y -> 10**y. This is another reason why it is good to separately convert metallicity for gas phase and stars (see my point 1).

bwvdnbro:
I think I have implemented all the changes you requested.

EvgeniiChaikin (approved):

EvgeniiChaikin:
Thanks! It all looks good now! Here are some example

stellar_mass_gas_sf_metallicity_total_lom_30
stellar_mass_star_fe_over_h_lom_50

PR #133, opened 2022-09-01
Add terrazas 2017 BH mass vs. SFR data

MatthieuSchaller:
Will be interesting to compare how rapidly our model(s) quench via AGN


MatthieuSchaller:
I'll add the plots to the pipeline next.

data/BlackHoleMassStarFormationRate/conversion/convertTerrazas2017.py
```diff
@@ -0,0 +1,81 @@
...
+BH_mass = unyt.unyt_array(10 ** data[:, 3], units="Solar_Mass")
+BH_mass_hi = unyt.unyt_array(10 ** (data[:, 3] + data[:, 4]), units="Solar_Mass")
+BH_mass_lo = unyt.unyt_array(10 ** (data[:, 3] - data[:, 4]), units="Solar_Mass")
+BH_mass_scatter = unyt.unyt_array([BH_mass_lo, BH_mass_hi])
+
+# Star formation rates (uncertainty taken from Sec. 2)
+SFR = unyt.unyt_array(10 ** data[:, 1], units="Solar_Mass / year")
+SFR_hi = unyt.unyt_array(10 ** (data[:, 1] + 0.3), units="Solar_Mass / year")
+SFR_lo = unyt.unyt_array(10 ** (data[:, 1] - 0.3), units="Solar_Mass / year")
+SFR_scatter = unyt.unyt_array([SFR_lo, SFR_hi])
```
bwvdnbro, line 39: You need something like
```
SFR_scatter = unyt.unyt_array([SFR - SFR_lo, SFR_hi - SFR])
```
The scatter is the actual error on the values and not the lower and upper bounds.

data/BlackHoleMassStarFormationRate/conversion/convertTerrazas2017.py
```diff
@@ -0,0 +1,81 @@
...
+data = np.loadtxt(input_filename)
+
+# Stellar masses
+Mstar = unyt.unyt_array(10 ** data[:, 0], units="Solar_Mass")
+
+# Black Hole Masses
+BH_mass = unyt.unyt_array(10 ** data[:, 3], units="Solar_Mass")
+BH_mass_hi = unyt.unyt_array(10 ** (data[:, 3] + data[:, 4]), units="Solar_Mass")
+BH_mass_lo = unyt.unyt_array(10 ** (data[:, 3] - data[:, 4]), units="Solar_Mass")
+BH_mass_scatter = unyt.unyt_array([BH_mass_lo, BH_mass_hi])
```
bwvdnbro, line 33: Same here.

bwvdnbro (changes requested):
I think the scatter is provided in the wrong way: you need to provide the error and not the upper and lower limit.

With the current values of scatter, the data looks like this:
Terrazas_before

While I think it should actually look like this:
Terrazas_after

The script `plot_individual_dataset.py` is quite useful for visual inspection of the newly added data.

MatthieuSchaller:
Thanks. I think I will never remember correctly how matplotlib's errobar function works. Somehow the choice they made feels non-intuituve but by now I should know better.

bwvdnbro:
For me this is good to go. I don't know if you want to wait for Evgenii's review before merging this?

MatthieuSchaller:
We can go ahead. I'll add another review on the plotting side.

PR #135, opened 2022-09-09
Add the Thorne+21 high-z GSMF data (DEVILS survey)

MatthieuSchaller:


data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
```diff
@@ -0,0 +1,109 @@
...
+citation = "Thorne et al. (2021)"
+bibcode = "2021MNRAS.505..540T"
+name = "Galaxy Stellar Mass Function"
+plot_as = "line"
+h_obs = 0.6777
+h = cosmology.h
+
+comment = (
+    f"GSMF at redshifts {min(redshifts)}<z{max(redshifts)}, "
+    "from the DEVILS suvery (D10-COSMOS field)."
```
EvgeniiChaikin, line 55:
```suggestion
    "from the DEVILS survey (D10-COSMOS field)."
```

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
```diff
@@ -0,0 +1,109 @@
...
+
+citation = "Thorne et al. (2021)"
+bibcode = "2021MNRAS.505..540T"
+name = "Galaxy Stellar Mass Function"
+plot_as = "line"
+h_obs = 0.6777
+h = cosmology.h
+
+comment = (
+    f"GSMF at redshifts {min(redshifts)}<z{max(redshifts)}, "
```
EvgeniiChaikin, line 54:
```suggestion
    f"GSMF at redshifts {min(redshifts)}<z<{max(redshifts)}, "
```

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
```diff
@@ -0,0 +1,109 @@
...
+multi_z.associate_comment(comment)
+multi_z.associate_cosmology(cosmology)
+multi_z.associate_maximum_number_of_returns(1)
+
+for z in redshifts:
+    tag = str(z).replace(".", "p")
+
+    processed = ObservationalData()
+
+    redshift = z
```
EvgeniiChaikin, line 73: I don't understand the meaning of this line

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
```diff
@@ -0,0 +1,109 @@
...
+    raw_filename = f"../raw/Thorne2021_z{tag}.txt"
+
+    M, N, sigma = np.loadtxt(raw_filename, delimiter=" ").T
+
+    mass = unyt.unyt_array(M, units=unyt.Solar_Mass)
```
EvgeniiChaikin, line 78: Here you want
```suggestion
    mass = unyt.unyt_array(10 ** M, units=unyt.Solar_Mass)
```

EvgeniiChaikin (changes requested):
Thanks @MatthieuSchaller !

Besides small comments in the code, I have two questions,

- Where does the raw data come from? Is it private communication or I can find it in the paper?

- I tried plotting the data at several redshirts. At `z=2.5` everything looks fine but at `z=0` the error bars are suspiciously large and the data itself is a bit off compared to GSMFs from the other surveys (see the plots below)

adaptive_stellar_mass_function_50_z2p5

adaptive_stellar_mass_function_50_z0p0

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
MatthieuSchaller: Fixed.

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
MatthieuSchaller: Fixed.

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
MatthieuSchaller: Fixed.

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+redshifts = [
+    0.064,
+    0.12,
+    0.18,
+    0.25,
+    0.32,
+    0.41,
+    0.51,
```
EvgeniiChaikin, line 32: If the data looks bad at `z=0` (compared to the other data we have at these z), would it be better to not show it at all at these low `z`? This can be realised, e.g., by commenting out a few first numbers from this list.

EvgeniiChaikin (approved):

data/GalaxyStellarMassFunction/conversion/convertThorne2021.py
MatthieuSchaller: Can we rather choose the range to use in the pipeline itself?

EvgeniiChaikin: I am not sure. Currently, there is no such functionality. Here we store Thorne's data as a _single_ file that has GSMFs at _multiple_ z. When we add this data to the plot, the pipeline will always try to find Thorne's GSMF that is closest to the snapshot's redshift. So if we make our plots at `z=0`,  the pipeline will use Thorne at `z=0.064`


If we would like to exclude certain redshifts in a multi-z file at the moment when we make plots, we would need to make a change in the velociraptor-python library. Do you think it would be worth it?

MatthieuSchaller: In principle, yes, as there is nothing fundamentally wrong with the low-z data here. 

But, that's a big change so don't think we should embark in that. So we could apply your proposed solution here.
From Fig. 9 of https://arxiv.org/pdf/2011.13605.pdf, I propose we keep the data from z>0.2.
What do you think?

EvgeniiChaikin: We could comment out the first two numbers, `0.064` and `0.12`, and change Delta z from  `0.25` to `0.15` in 

`processed.associate_redshift(z, redshift_lower=z - 0.25, redshift_upper=z + 0.25)`

MatthieuSchaller: I have changed that to use something slightly more dynamic. Let me know what you think.

Also, I have cancelled the first three outputs.

EvgeniiChaikin: I am happy with the current version. Let's merge it! 

PS Here's the `z=0` plot after today's changes

adaptive_stellar_mass_function_50

MatthieuSchaller: Wait, that's not great. Ideally that dataset should no appear at all at z=0.

MatthieuSchaller: I put a limit of 0.2 for the lowest bin, should I not expect the code to honour that?

PR #137, opened 2022-09-23
Add High-Z UVLF

JBorrow:
Adds UVLF at z~5-6 from Rosdahl (simulation) and Bouwens (obs) data. Depends on https://github.com/SWIFTSIM/velociraptor-comparison-data/pull/136 because I did this wrong.

### DIFF
--- a/data/GalaxyLuminosityFunction/conversion/convertBouwens2017.py
+++ b/data/GalaxyLuminosityFunction/conversion/convertBouwens2017.py
@@ -1,0 +1,124 @@
+from velociraptor.observations.objects import ObservationalData
+
+import unyt
+import os
+import sys
+import csv
+import numpy as np
+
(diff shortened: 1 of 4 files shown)


data/GalaxyLuminosityFunction/conversion/convertBouwens2017.py
```diff
@@ -0,0 +1,128 @@
...
+)
+
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Meta-data
+comment = (
+    "UV luminosity function from Hubble Frontier Fields, at z=6. "
+    "Data obtained assuming h=0.7, from Table 5."
```
MatthieuSchaller, line 99: Might be good to add something here about what dust correction and maybe volume correction they may had added.

MatthieuSchaller:
Any updates?

PR #143, opened 2022-10-27
Add M-TNG GSMF and M* - Mhalo relations

MatthieuSchaller:


data/GalaxyStellarMassHaloMass/conversion/convertMTNG.py
```diff
@@ -0,0 +1,67 @@
...
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+# Cosmology
+h_sim = cosmology.h
+Omega_b = cosmology.Ob0
+Omega_m = cosmology.Om0
```
EvgeniiChaikin, line 15: Don't we want to use here the values taken directry from the MTNG paper, `Omega_m = 0.3089, Omega_b = 0.0486`?

EvgeniiChaikin (approved):
Thank you @MatthieuSchaller!

Both curves look very nice! I have only one small comment in the code.

stellar_mass_halo_mass_MBN98_centrals_ratio_30

adaptive_stellar_mass_function_50

data/GalaxyStellarMassHaloMass/conversion/convertMTNG.py
MatthieuSchaller: Good point. I have made that change.

MatthieuSchaller:
@EvgeniiChaikin btw, the Mstar - Mhalo data is using M_200 and not M_BN98 so the plot above is not quite correct.

EvgeniiChaikin:
@MatthieuSchaller yes, you are right. We can then include these new data in our plots that show M_200 rather than M_BN98.

PR #144, opened 2022-11-02
Add stellar mass - BH mass relation from Graham 2022

EvgeniiChaikin:
Add stellar mass - black hole mass relation from Graham2022 (https://arxiv.org/pdf/2209.14526.pdf).

stellar_mass_black_hole_mass_50

Note that

Graham assumes a Kroupa IMF while we use the Chabrier IMF. On page 11 they say "Bernardi et al. (2010) suggest a reduction of just 0.05 dex to the logarithm of Υ∗ (≡ M∗/L) to convert from the Chabrier (2003) IMF to the Kroupa (2002)". 

The script includes the 0.05 dex correction to galaxy stellar masses.


bwvdnbro (approved):
Looks good!

I was a bit surprised the `±` does not cause issues, so I looked into it some more. Seems that Python natively support UTF-8 and `±` is a valid UTF-8 character. So all good on that front.

PR #149, opened 2022-11-30
Add data for metallicity plots

correac:
I want to add two datasets to the stellar metallicity-stellar mass plots: Zahid et al. (2017) and Kudritzki et al. (2016).


data/GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py
```diff
@@ -0,0 +1,62 @@
...
+Z_median = 10 ** raw[:, 1] * unyt.dimensionless * Z_solar_obs / solar_metallicity
+
+
+# Meta-data
+comment = (
+    "Data obtained from Kudritzki et al. (2016). No need for "
+    "h-correction. The metallicity is expressed in units of solar metallicity, using Z=0.0142. "
+    f"This has been corrected to use Z_solar={solar_metallicity}. "
+)
+citation = "Kudritzki et al. (2016) (SDSS)"
```
EvgeniiChaikin, line 35: I do not understand why it says SDSS here. Doesn't Kudritzki et al. (2016) compare only 15 local galaxies?

data/GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py
```diff
@@ -0,0 +1,62 @@
...
+# Read the data
+raw = np.loadtxt(input_filename)
+M_star = 10 ** raw[:, 0] * unyt.Solar_Mass
+Z_median = 10 ** raw[:, 1] * unyt.dimensionless * Z_solar_obs / solar_metallicity
+
+
+# Meta-data
+comment = (
+    "Data obtained from Kudritzki et al. (2016). No need for "
+    "h-correction. The metallicity is expressed in units of solar metallicity, using Z=0.0142. "
```
EvgeniiChaikin, line 32: You could please point me to the text in the paper where Kudritzki et al. (2016) say that the value of Zsolar=0.0142 is used?

I couldn't find this value in the text.

data/GalaxyStellarMassStellarMetallicity/raw/Kudritzki_2016.txt
```diff
@@ -0,0 +1,18 @@
+# mass-metallicity data from Kudritzki et al. (2016) (fig. 9, arxiv: 1607.04325)
```
EvgeniiChaikin, line 1: I have noticed that Fig. 9 from Kudritzki et al. (2016) has 15 galaxies but this table has 14.

EvgeniiChaikin (changes requested):
Hi @correac ,

thank you for the update!

The code looks ready and the plots with the new data look great!

I have left only three small comments, which are all about `Kudritzki et al. (2016)`.

data/GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py
correac: That's right. I will remove SDSS.

data/GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py
correac: I couldn't find it either. They mention they adopt 12 + log(O/H)solar=8.69 from Asplund+ (2009), which has Zsun=0.0142. In Zahid+2017 (with Kudritzki as second author) they specifically say they adopt solar abundances as in Asplund+ (2009) using Z=0.0142. From this I assume Kudritzki uses Zsolar = 0.0142.

data/GalaxyStellarMassStellarMetallicity/raw/Kudritzki_2016.txt
correac: Let me check, perhaps I missed one.

data/GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py
EvgeniiChaikin: Note that in COLIBRE we also use the metallicities from Asplund+ (2009) but our value is 0.0134 rather than 0.0142. 

Screenshot from 2022-12-01 10-50-02

correac: Yes that's right. But given that Zahid+17 used Z=0.0142. I assumed this was also used in Kudritzki+. The difference between Zsolar=0.0142 and Zsolar(calibre)=0.0134 was considered in the script.

EvgeniiChaikin: I found two papers of Kudritzki+ where they say what solar metallicity is used. It indeed comes from  Asplund+ (2009) but the value is Zsolar=0.014 instead of 0.0142. See https://arxiv.org/pdf/1404.7244.pdf (page 10) and https://arxiv.org/pdf/2102.04135.pdf (page 4).

If you agree, in the Kudritzki+ conversion script we could change Z=0.0142 to Z=0.014 and say in the description that this value comes from Asplund+ (2009).

correac: Right. I would prefer to keep it as Z=0.0142, to follow from Zahid+ comments. But if you want it Z=0.014, we can change it :) 

Note that in the script I am doing the conversion from Zsolar = 0.0142 to Zsolar = 0.0134 (also in Zahid).

EvgeniiChaikin: I think either is fine because we are talking about very small differences :)

EvgeniiChaikin (approved):

PR #150, opened 2022-12-12
Use the Asplund09 value for Zsolal and update references to Fraser-McKelvie's data 

EvgeniiChaikin:
- Use the Asplund09 value for solar metallicity in `cosmology.py` (this is the value that we should use everywhere in the code; I am not sure why in `cosmology.py` we currently have a different value).
- Update references to Fraser-McKelvie's data because the paper has been published (before we were referring to the submitted but not final version. Also indicate in the data's legend that the data comes from the SAMI survey).

### DIFF
--- a/cosmology.py
+++ b/cosmology.py
@@ -1,6 +1,7 @@
 from astropy.cosmology import Planck15 as cosmology
 
-solar_metallicity = 0.0126
+# Asplund et al. (2009)
+solar_metallicity = 0.0134
 
 # IMF conversion factors from Furlong+ 2015
 kroupa_to_chabrier_mass = 0.912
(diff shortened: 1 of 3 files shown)


bwvdnbro:
A quick `grep` shows that there are four other scripts that use the `solar_metallicity` variable defined in `cosmology.py`:
 - `GalaxyStellarMassStellarMetallicity/conversion/convertKudritzki2016.py`
 - `GalaxyStellarMassStellarMetallicity/conversion/convertZahid2017.py`
 - `GalaxyStellarMassStellarMetallicity/conversion/convertGallazzi2005.py`
 - `GalaxyStellarMassAlphatoIron/conversion/convertGallazzi2021.py` (although this one does not actually use the value, it only writes it to the description)

Are we happy with the changes this will cause there?

correac:
The changes look good @EvgeniiChaikin !

EvgeniiChaikin:
Thanks @correac for taking a look!

@bwvdnbro everything looks good to me too!

bwvdnbro (approved):

PR #154, opened 2023-02-24
SNIa data update

Fonotec:
This includes 14 more observations for the cosmic SNIa rate and one observation for the specific SFR vs the specific SNIa rate.


MatthieuSchaller:
What about the SNIa rate files? Shouldn't these also be converted?

Fonotec:
Yes, this includes the conversion files

Fonotec:
Btw it would be good if you merge this because then it can be included in the pipeline, but as long as you don't merge this I will add more SNIa data if I find more datasets.

MatthieuSchaller (approved):
Looks good to me.

PR #156, opened 2023-03-23
Update Graham & Sahu (2022) reference

EvgeniiChaikin:
Update the reference to Graham & Sahu data (2022) because the paper has been published. The previous version was an arXiv version.


MatthieuSchaller:
Pressed "OK" too early... I was about to ask whether the data had changed between the arxiv and published version?

EvgeniiChaikin:
I did a cross-check of the values of all fields from their table 1 for some of the galaxies, I did not find any differences between the published version and the arXiv one.

MatthieuSchaller:
Great! Thanks for confirming.

PR #157, opened 2023-04-03
Fix conversion script for Bauer2013+, all galaxies

EvgeniiChaikin:
The older version would select the lower limits on the detected SFRs. The new version correctly selects the data for all galaxies.

**Before**

stellar_mass_specific_sfr_all_50_old

**After**

stellar_mass_specific_sfr_all_50


EvgeniiChaikin:
@MatthieuSchaller Is there anything else that you think we need to do in this pull request?

MatthieuSchaller (approved):

PR #159, opened 2023-04-16
Add swiftsimio to the requirements in order to fix the error in the KS-relation scripts

EvgeniiChaikin:
The error would manifest if one would try to install this repository from scratch (as is done in the automated tests), without having `swiftsimio` installed.

Here is an example of this error in the automated tests

Screenshot from 2023-04-16 15-33-39

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ velociraptor
 matplotlib
 numpy
 scipy
+swiftsimio


MatthieuSchaller:
I don't like that this script requires swiftsimio. Any reason we need to pull SWIFT-related things into this?

EvgeniiChaikin:
Here is the idea behind this problem.

The implementation of the SOAP catalogue in the `velociraptor-python` library allows usage of swift's most generic cosmology. Because of this, in order to always be consistent (unit-wise and structure-wise) with swift, the `velociraptor-python` library imports certain functionality directly from the `swiftsimio` library. You can see this code here (https://github.com/SWIFTSIM/velociraptor-python/blob/master/velociraptor/catalogue/soap_catalogue.py#L15)

The next question is: how is this related to the problem in the current PR given that in the `velociraptor-comparison-data` package, we only work with the _observational-data_ submodule of the `velociraptor-python` library? 

The answer is that when in some of the conversion scripts here, we do the import as

` from velociraptor.observations.object import ObservationalData`,

python is going to initialise each submodule in the hierarchy by calling the corresponding `__init__.py` file.  The  `__init__.py` file of the `velociraptor` submodule contains the line

`from velociraptor.catalogue.soap_catalogue import SOAPCatalogue`,

which imports the SOAP catalogue class. This causes the error that this PR is trying to fix.

A solution to this problem without introducing the explicit dependency of this package on the `swiftsimio` library could be to simply move the lines

`from velociraptor.catalogue.soap_catalogue import SOAPCatalogue`
and
`from velociraptor.catalogue.velociraptor_catalogue import VelociraptorCatalogue`

from the global scope inside the function `load()` (see the file https://github.com/SWIFTSIM/velociraptor-python/blob/master/velociraptor/__init__.py#L32). Then the lines won't be executed when the `__init__.py` file is called

@MatthieuSchaller  what do you think?

MatthieuSchaller:
I see the logic of all this, yes. I am just wondering why we would need to pass in a generic cosmology object all the way.

EvgeniiChaikin:
To avoid rewriting (duplicating) the code for the generic cosmology class that already exists in the `swiftsimio` library.

MatthieuSchaller:
Sure but the problem with the solution here is that we are now tying the data to SWIFT even though that has nothing to do with it.

MatthieuSchaller:
Yes, that sounds better.

MatthieuSchaller (approved):

PR #160, opened 2023-04-16
Add Davies et al. (2016) data for sSFR vs stellar mass (GAMA, SF galaxies)

EvgeniiChaikin:
stellar_mass_specific_sfr_active_50

### DIFF
--- a/data/GalaxyStellarMassSpecificStarFormationRate/conversion/convertDavies2016.py
+++ b/data/GalaxyStellarMassSpecificStarFormationRate/conversion/convertDavies2016.py
@@ -1,0 +1,138 @@
+from velociraptor.observations.objects import (
+    ObservationalData,
+    MultiRedshiftObservationalData,
+)
+
+import unyt
+import numpy as np
+import os
+import sys
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+
+def SFR_vs_Mstar_relation(norm, norm_err, slope, slope_err, log10Mstar):
+    """
+    norm: normalization of the fitting relation in units of Msun/yr (in log-log space)
+    norm_err: error on the normalization
+    slope: slope of the fitting relation (in log-log space)
+    slope_err: error on the slope
+    log10Mstar: log10 of stellar mass in units of Msun
+
+    returns: SFR, positive and negative errors on the SFR, all in units of Msun/yr
+    """
+
+    log10SFR = slope * (log10Mstar - 10.0) + norm
+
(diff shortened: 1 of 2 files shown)


data/GalaxyStellarMassSpecificStarFormationRate/conversion/convertDavies2016.py
```diff
@@ -0,0 +1,138 @@
...
+name = "Galaxy Stellar Mass - Galaxy sSFR"
+comment = (
+    "Assumes Chabrier (2003) IMF. "
+    "Uses GAMA II equatorial dataset with redshifts from z=0 to 0.35. "
+    "Plots fitting relations to the actual data in four redshift bins, "
+    " which are defined using a normalization and a slope. "
+    "The fits correspond to active galaxies only. Star-forming galaxies are defined based on a u−r cut of 1.4."
+    "The fitting range is valid approximately within 9.25 < log10[M*/Msun] < 11. "
+    "No h-correction is required. "
+    "Cosmology: H0 = 70 kms-1 Mpc-1, Omega_lambda = 0.7 and Omega_m = 0.3."
```
MatthieuSchaller, line 65: Don't agree that correction is needed.  The correction to apply is even given in the text file.

EvgeniiChaikin: The correction for h is given in the entry where they provide values of cosmic star formation rate density. We, however, only plot Mstar vs sSFR. I.e. this script does not plot anything that would have volume-related units.

EvgeniiChaikin: I had a closer look and, apparently, for Mstar the h correction is needed.

The original paper refers to `Taylor, E. N., Hopkins, A. M., Baldry, I. K., et al. 2011, MNRAS, 418, 1587` to tell us how Mstar was calculated. Based on eq. 3 from Taylor et al. 2011, I figured that the correction should be `Mstar -> Mstar (h_sim / h_obs) ** -2`.

EvgeniiChaikin: I will now also check whether anything extra needs to be added for the SFR.

EvgeniiChaikin:
It is not fully clear to me how to correct for h in the sSFR because sSFR depends slightly _non-linearly_ on luminosity.

I think the safest option is not to do any h-corrections for the sSFR at all. This is because sSFR = SFR / Mstar, and Mstar and SFR depend on h approximately in the same way (through the dependence on the luminosity distance squared).

PR #162, opened 2023-04-24
Add Stellar and baryon fraction data

Moyoxkit:
Adds the stellar and baryon fraction data from Gonzalez2013, Zhang2011 and Kravtsov2018

Note there are some IMF conversion factors in some of the files, the values of these conversions are taken from Chiu2018.


data/HaloMassBaryonFractions/conversion/convertGonzalez2013.py
```diff
@@ -0,0 +1,168 @@
...
+FLATCDM = FlatLambdaCDM(H0=cosmology.H0, Om0=cosmology.Om0)
+
+output_filename = "Gonzalez2013.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Cosmology correction factor
+correction_factor = h_sim / 70.2
```
EvgeniiChaikin, line 26:
```suggestion
correction_factor = h_sim / 0.702
```

data/HaloMassBaryonFractions/conversion/convertGonzalez2013.py
```diff
@@ -0,0 +1,168 @@
...
+
+# Data from table 2 from the paper, fstar includes the change to the mass due to IMF change as documented by Chiu+18
```
EvgeniiChaikin, line 28: Can you please be more precise about the change in the IMF?

What is the final IMF after the conversion? Chabrier (2003)? If so, please say it in the comment.

data/HaloMassBaryonFractions/conversion/convertZhang2011.py
```diff
@@ -0,0 +1,253 @@
...
+error_fb_500_p = error_fb_500_p / (Omega_b / Omega_m)
+error_fb_500_m = error_fb_500_m / (Omega_b / Omega_m)
+
+# Define the scatter as offset from the mean value
+x_scatter = unyt.unyt_array((M_500_err, M_500_err))
+y_scatter = unyt.unyt_array((error_fb_500_m, error_fb_500_p))
+
+# Meta-data
+comment = "Baryon fraction data from the combined X-ray and optical observations"
+citation = "Zhang et al. (2011)"
```
EvgeniiChaikin, line 226: Are you sure these data are correct?

They have a Corrigendum to the original paper. See https://ui.adsabs.harvard.edu/abs/2012A%26A...544C...3Z/abstract)

data/HaloMassStellarFractions/conversion/convertGonzalez2013.py
```diff
@@ -0,0 +1,125 @@
...
+            0.041,
+            0.022,
+            0.023,
+            0.023,
+            0.022,
+            0.020,
+        ]
+    )
+    * correction_factor ** (-1.5)
+    * 0.76
```
EvgeniiChaikin, line 56: Where does this number of 0.76 come from?

If you use this constant several times throughout the file, please define it as a variable at the top of the file. And add a comment explaining why this correction factor is needed and how it has been computed.

data/HaloMassStellarFractions/conversion/convertKravtsov2013.py
```diff
@@ -0,0 +1,95 @@
...
+FLATCDM = FlatLambdaCDM(H0=cosmology.H0, Om0=cosmology.Om0)
+
+output_filename = "Kravtsov2018.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Cosmology correction factor
+correction_factor = h_sim / 70.0
```
EvgeniiChaikin, line 26: You probably want
```suggestion
correction_factor = h_sim / 0.7
```

data/HaloMassStellarFractions/conversion/convertZhang2011.py
```diff
@@ -0,0 +1,185 @@
...
+FLATCDM = FlatLambdaCDM(H0=cosmology.H0, Om0=cosmology.Om0)
+
+output_filename = "Zhang2011.hdf5"
+output_directory = "../"
+
+if not os.path.exists(output_directory):
+    os.mkdir(output_directory)
+
+# Cosmology correction factor
+correction_factor = h_sim / 70.0
```
EvgeniiChaikin, line 26:
```suggestion
correction_factor = h_sim / 0.7
```

data/HaloMassBaryonFractions/conversion/convertGonzalez2013.py
```diff
@@ -0,0 +1,168 @@
...
+            0.013,
+        ]
+    )
+    * correction_factor ** (-1.5)
+    - 0.24 * f_starer
+)
+
+
+# Convert to proper units
+M_500 = unyt.unyt_array(M_500, units="Msun")
```
EvgeniiChaikin, line 124: The values of M_500 seem to be missing an ~1e13 factor. Please check it.

Please also check the normalisation of the Y-axis values.

I was not able to plot these data in my tests.

data/HaloMassStellarFractions/conversion/convertGonzalez2013.py
```diff
@@ -0,0 +1,125 @@
...
+            0.003,
+            0.003,
+        ]
+    )
+    * correction_factor ** (-1.5)
+    * 0.76
+)
+
+# Convert to proper units
+M_500 = unyt.unyt_array(M_500, units="Msun")
```
EvgeniiChaikin, line 81: The values of M_500 seem to be missing an ~1e13 factor. Please check it.

Please also check the normalisation of the Y-axis values.

I was not able to plot these data in my tests.

EvgeniiChaikin (changes requested):
Thank you a lot for the update @Moyoxkit!

It already looks good, but we need to improve it a little bit before the merge.

A few major comments:

- Please check the conversion in Gonzalez+2013 scripts. There is something wrong with the normalization of M_500; the values are too small. There could also be something wrong with the Y-axis values. I was not able to correctly visualize these data.

- Please move all numbers from the scripts to raw files. Otherwise, it is difficult to read the scripts; they become too long and it is easier to make a mistake. You can do it as follows:

If you have, say, three arrays
```
A = np.array([1,2,3,4])
B = np.array([-1,-2,-3,-4])
C = np.array([0.1, 0.1, 0.1 , 0.1])
```
Save them as, e.g.,
`np.savetxt("raw/AuthorYear.txt", np.vstack([A,B,C]).T, header="Variable X [Units], Variable Y [Units], Err Variable Y [Units]")`

and then replace the part of the code where you define the arrays with
```
raw_data = np.loadtxt("raw/AuthorYear.txt")
a = raw_data[:, 0]
b = raw_data[:, 1]
c = raw_data[:, 2]
```
Please specify in the header of the raw files, which units the original numbers are in

- I also left a few smaller comments in the scripts.

data/HaloMassBaryonFractions/conversion/convertZhang2011.py
EvgeniiChaikin: This is resolved now. No change is needed because @Moyoxkit was not using the stellar masses from table 1. Instead, @Moyoxkit uses stellar fractions from Table 2, which is not affected by the error.

EvgeniiChaikin (approved):

PR #168, opened 2023-05-19
Add Gas Metallicity vs. Stellar Mass for massitve galaxies at z=0.7 from Lewis et al (2023)

EvgeniiChaikin:
https://arxiv.org/abs/2304.12343

stellar_mass_gas_sf_metallicity_fromz_lom_50


EvgeniiChaikin:
It is not clear from the paper.

I can add an `(h/h_obs)**2` correction assuming that the stellar masses were derived as` Mstar \propto measured flux in a given band * (Luminosity distance)**2`

MatthieuSchaller:
They assumed h=0.7 so corrections would be small anyway. Let's maybe just list the assumed cosmology in the comment.

```
H0 = 70 km s−1 Mpc−1, ΩM = 0.3, and ΩΛ = 0.7.
```

MatthieuSchaller (approved):

PR #171, opened 2023-06-16
Add cosmic star formation rate history data from Cochrane+ 2023

EvgeniiChaikin:
**Plot from the paper:**

Screenshot from 2023-06-16 21-03-01

**Plot from the pipeline with the new data added**

star_formation_history


data/StarFormationRateHistory/conversion/convertCochrane2023.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Fetch the fields we need
+    z_m, z_p = data[:, 0], data[:, 1]
+    z = 0.5 * (z_m + z_p)
+    a = 1.0 / (1.0 + z)
+    a = unyt.unyt_array(a, units="dimensionless")
+
+    SFR, SFR_err = data[:, 7], data[:, 8]  # Msun / yr / Mpc**3
+    SFR_scatter = unyt.unyt_array((SFR_err, SFR_err), units="Msun/yr/Mpc**3")
+    SFR = unyt.unyt_array(SFR, units="Msun/yr/Mpc**3")
```
MatthieuSchaller, line 48: No cosmology correction here? e.g. volume changes because of h?

EvgeniiChaikin: The paper does seem to explain how to correctly account for h.

I would assume that SFR \propto h**-2 while volume \propto h**-3, so the SFR density should be proportional to h, meaning that the correction should be very small.

EvgeniiChaikin: Since I am not completely sure what is the right dependence on h, and the h-correction should be very small anyway, I would leave everything as is.

Alternatively, I could write an email to the authors.

What do you think is the best option?

MatthieuSchaller: Let's add then a message in the description saying that no correction was applied.

EvgeniiChaikin: Done

MatthieuSchaller (approved):

PR #176, opened 2023-08-21
Include scatter in GSMF from Leja+ 2020

EvgeniiChaikin:
I added scatter to the GSMF from Leja+ 2020.

Note that the code for the scatter was already present in the conversion script, I just had to propagate the scatter to the hdf5 file.

I reduced the number of points for the GSMF from 100 to 30, so that it better looks when plotted. Here is an example

GSMF_Leja_with_sigma

### DIFF
--- a/data/GalaxyStellarMassFunction/conversion/convertLeja2020.py
+++ b/data/GalaxyStellarMassFunction/conversion/convertLeja2020.py
@@ -17,7 +17,9 @@ z_list = [0.2, 0.5, 1, 1.5, 2, 3]
 
 output_filename = "Leja_2020.hdf5"
 output_directory = "../"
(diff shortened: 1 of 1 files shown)


MatthieuSchaller:
Do we want both the lines and the points with error bar then?

EvgeniiChaikin:
Good point!

We want to keep it as is, i.e. show the data in the `line' format, as opposed to `points' format. This is because the Leja+ 2020 data come from a model, meaning that it is a _continuous_ function of Mstar.

The plot I attached above does not fully reflect how the GSMF will look in the pipeline, because it comes from a separate python notebook that I wrote to test the changes in the script.

Here is how the GSMF will look in the pipeline if we merge this PR as is

adaptive_stellar_mass_function_50_calibration

I think it looks pretty good, so I'd not change anything. What do you think?

MatthieuSchaller:
That looks fine like this, yes.

Accepted.

MatthieuSchaller (approved):

PR #189, opened 2023-12-15
Add metallicity data from Chartab+23, Sanders+23, Pharo+23

robjmcgibbon:
stellar_mass_gas_sf_metallicity_mol_hifloor_30
stellar_mass_gas_sf_metallicity_mol_lowfloor_50
stellar_mass_star_fe_over_h_mol_hifloor_50

### DIFF
--- a/data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
+++ b/data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
@@ -28,7 +28,7 @@ name = "Galaxy Stellar Mass - Gas Metallicity"
 plot_as = "points"
 redshift = 0.7
 
-all_indicators = ['Ne3O2', 'O3HB', 'O3O2', 'R23']
+all_indicators = ["Ne3O2", "O3HB", "O3O2", "R23"]
 indicators, Mstar, Zgas, Zgas_std = [], [], [], []
 
 with open(input_filename, "r") as file:
@@ -37,7 +37,7 @@ with open(input_filename, "r") as file:
         if c > 2:
             m = float(row[1]) + 0.03 * (all_indicators.index(row[0]) - 1.5)
             indicators.append(row[0])
-            Mstar.append(10**m * (h_sim / h_obs) ** -2)
+            Mstar.append(10 ** m * (h_sim / h_obs) ** -2)
             Zgas.append(float(row[2]))
             Zgas_std.append(float(row[3]))
 
@@ -50,10 +50,7 @@ for indicator in all_indicators:
 
     mask = np.array(indicators) == indicator
     processed.associate_x(
-        Mstar[mask], 
-        scatter=None, 
-        comoving=False, 
-        description="Galaxy Stellar Mass"
+        Mstar[mask], scatter=None, comoving=False, description="Galaxy Stellar Mass"
     )
     processed.associate_y(
         Zgas[mask],
@@ -75,4 +72,3 @@ for indicator in all_indicators:
         os.remove(output_path)
 
     processed.write(filename=output_path)
-
--- a/data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
+++ b/data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
@@ -1,0 +1,78 @@
+from velociraptor.observations.objects import ObservationalData
+
+import csv
+import numpy as np
+import os
+import sys
+import unyt
+
+# Exec the master cosmology file passed as first argument
+with open(sys.argv[1], "r") as handle:
+    exec(handle.read())
+
+h_obs = 0.7
+h_sim = cosmology.h
+
+input_filename = f"../raw/Pharo2023.csv"
+
+comment = (
+    "The relationship between stellar mass and gas-phase metallicity "
+    "at 0.3<z<85 from the HALO7D survey. "
+    "Uses Kroupa initial mass function and h=0.7. "
+    f"h-corrected for SWIFT using cosmology: {cosmology.name}. "
+    "The metallicity is calculated using 4 different methods. "
+    "The stellar mass for each method is offset by 0.03 dex "
+)
+bibcode = "2023ApJ...959...48P"
+name = "Galaxy Stellar Mass - Gas Metallicity"
+plot_as = "points"
+redshift = 0.7
+
+all_indicators = ['Ne3O2', 'O3HB', 'O3O2', 'R23']
+indicators, Mstar, Zgas, Zgas_std = [], [], [], []
+
+with open(input_filename, "r") as file:
+    data = csv.reader(file, delimiter=",")
+    for c, row in enumerate(data):
+        if c > 2:
+            m = float(row[1]) + 0.03 * (all_indicators.index(row[0]) - 1.5)
+            indicators.append(row[0])
+            Mstar.append(10**m * (h_sim / h_obs) ** -2)
+            Zgas.append(float(row[2]))
+            Zgas_std.append(float(row[3]))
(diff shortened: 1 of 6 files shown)


data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
```diff
@@ -0,0 +1,74 @@
...
+plot_as = "points"
+redshift = 0.7
+
+all_indicators = ["Ne3O2", "O3HB", "O3O2", "R23"]
+indicators, Mstar, Zgas, Zgas_std = [], [], [], []
+
+with open(input_filename, "r") as file:
+    data = csv.reader(file, delimiter=",")
+    for c, row in enumerate(data):
+        if c > 2:
```
EvgeniiChaikin, line 37: Please add a short comment, something like `# skip header', so that it is immediately clear to everyone why you are keeping the first rows.

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
```diff
@@ -0,0 +1,74 @@
...
+            m = float(row[1]) + 0.03 * (all_indicators.index(row[0]) - 1.5)
```
EvgeniiChaikin, line 38: I don't understand why you are adding a 0.03 dex offset.

I also don't understand the factor `(all_indicators.index(row[0]) - 1.5)` does preciesly

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
```diff
@@ -0,0 +1,74 @@
...
+            indicators.append(row[0])
+            Mstar.append(10 ** m * (h_sim / h_obs) ** -2)
```
EvgeniiChaikin, line 40: Are you converting the stellar masses from Kroupa IMF to Chabrier IMF? (there is a conversion factor stored in `cosmology.py`)

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
```diff
@@ -0,0 +1,74 @@
...
+    exec(handle.read())
+
+h_obs = 0.7
+h_sim = cosmology.h
+
+input_filename = f"../raw/Pharo2023.csv"
+
+comment = (
+    "The relationship between stellar mass and gas-phase metallicity "
+    "at 0.3<z<85 from the HALO7D survey. "
```
EvgeniiChaikin, line 20: do you mean `0.85` in place of `85`?

data/GalaxyStellarMassGasMetallicity/conversion/convertSanders2021.py
```diff
@@ -0,0 +1,71 @@
...
+        description="Galaxy Gas Metallicity",
+    )
+    processed.associate_citation(citation, bibcode)
+    processed.associate_name(name)
+    processed.associate_comment(comment)
+    processed.associate_redshift(redshift)
+    processed.associate_plot_as(plot_as)
+    processed.associate_cosmology(cosmology)
+
+    output_path = f"../Sanders2021_z{redshift}.hdf5"
```
EvgeniiChaikin, line 66: I'd write
```suggestion
    output_path = f"../Sanders2021_z{redshift:.1f}.hdf5"
```
to make formatting more explicit.

data/GalaxyStellarMassStellarMetallicity/raw/Chartab2023.txt
```diff
@@ -0,0 +1,12 @@
+# The relationship between stellar mass and stellar metallicity from Chartab et al. (2023)
+# The raw data is taken directly from Table 3 of the paper
+# log10 Mstar/Msun, log10 Zstar/Zsun, Zstar positive error, Zstar negative error
```
EvgeniiChaikin, line 3: Do they use the same solar metallicity as us? If not, we'd need to convert their values for stellar metallicity.

We use the solar values from Asplund et al. (2009)
```
Fe_H_Sun_Asplund = 7.5
Z_Sun_Asplund = 0.0134
```

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
robjmcgibbon: The offset is there just to prevent all the points ending up on top of each other. This can be seen from the second plot I have in the description of the PR. Joop ok'd this

`(all_indicators.index(row[0]) - 1.5)` gives an offset of `-1.5, -0.5, 0.5, 1.5` for `["Ne3O2", "O3HB", "O3O2", "R23"]` respectively

robjmcgibbon: The `-1.5` is just so that the offsets are centered on the original x value

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
robjmcgibbon: Updated

data/GalaxyStellarMassGasMetallicity/conversion/convertSanders2021.py
robjmcgibbon: Updated

data/GalaxyStellarMassStellarMetallicity/raw/Chartab2023.txt
robjmcgibbon: Yep they use Asplund, it's mentioned at the very end of section 1

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
robjmcgibbon: Screenshot from 2024-03-18 16-29-02
No sure why, but Joop said no conversion was needed

data/GalaxyStellarMassStellarMetallicity/raw/Chartab2023.txt
EvgeniiChaikin: Okay, good! Can we please just mention it somewhere in the script and/or raw file with the data, so that we don't forget why we perform no conversion?

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
EvgeniiChaikin: Probably because the conversion factor from Kroupa to Chabrier is close to unity, so it has only a small effect.

data/GalaxyStellarMassGasMetallicity/conversion/convertPharo2023.py
EvgeniiChaikin: Oh, I see now, thanks!

Do you think it'd be good to say in the comment to the data what the purpose of the offset is? (if I understood you correctly, this is done only to have a better visual representation of the data, ensuring that the x values of the different methods's data don't overlap)

robjmcgibbon: Yep, I've added a comment to describe the reason for the offset

data/GalaxyStellarMassStellarMetallicity/raw/Chartab2023.txt
robjmcgibbon: I added a comment to the raw data

EvgeniiChaikin:
Thanks, @robjmcgibbon , all looks good!

EvgeniiChaikin (approved):

PR #191, opened 2024-02-21
Update S. Driver's 2022 GSMF

EvgeniiChaikin:
1. Change year from 2021 to 2022 (2022 is the year of the publication)
2. Update the correction coefficient that moves the data exactly to z=0
3. Add the DOI of the GAMA4 paper to the conversion script
4. Use the correct number of the table from which the data are taken (changed it from 5 to 6)

**Regarding point two, the difference is less than 1 per cent**

Driver_check

**Regarding point four, here is a screenshot of Table 6 from the published paper**

Screenshot from 2024-02-20 15-13-34


MatthieuSchaller (approved):
Approved.

PR #192, opened 2024-03-20
Add data from Koprowski+24

robjmcgibbon:
All galaxies
stellar_mass_sfr_50

Active galaxies only
stellar_mass_sfr_active_50


data/GalaxyStellarMassStarFormationRate/conversion/convertKoprowski2024.py
```diff
@@ -0,0 +1,86 @@
...
+            (2.75, 3.75),
+            (3.75, 4.75),
+            (4.75, 6.75),
+        ]
+    ):
+        zmean = 0.5 * (zmin + zmax)
+        # Select all or star-forming galaxies
+        mask = raw[:, 0] == star_forming
+        # Select mass ranges with data at this redshift
+        mask &= raw[:, 3 + 2 * i] != 0
```
EvgeniiChaikin, line 50: Aren't there too many actions concentrated in this line?

Wouldn't it be clearer to first check the data against zeroes (let's call this mask one) and then, in the next line, do `np.logican_and(...)` between this mask and the star formation mask?

data/GalaxyStellarMassStarFormationRate/conversion/convertKoprowski2024.py
```diff
@@ -0,0 +1,86 @@
...
+bibcode = "2024arXiv240306575K"
+name = "Galaxy Stellar Mass-Star Formation Rate "
+plot_as = "line"
+
+for star_forming in [0, 1]:
+    multi_z = MultiRedshiftObservationalData()
+    multi_z.associate_citation(citation, bibcode)
+    multi_z.associate_name(name)
+    multi_z.associate_comment(comment)
+    multi_z.associate_cosmology(cosmology)
```
EvgeniiChaikin, line 33: Let's add also `multi_z.associate_maximum_number_of_returns(1)` here. This line will ensure that in the pipeline, only one redshift of these data (the one that is closest to the snapshot's redshift), will be shown.

robjmcgibbon: Thanks, that's a very useful feature I didn't know about. I've added `zmin` and `zmax` as arguments to `processed.associate_redshift`, which I hadn't included previously since they were causing too many lines to be shown.

data/GalaxyStellarMassStarFormationRate/conversion/convertKoprowski2024.py
robjmcgibbon: I've split it into two different lines

EvgeniiChaikin (approved):
All good! Note that before merging, I rebased this branch onto the most recent version of the master branch, which was updated yesterday.

PR #199, opened 2024-05-08
Add column density data from Oliver2024

robjmcgibbon:
Adding data from 4 different observational papers. This was compiled and provided to us by Oliver Thompson, and is all in the same format, so there is only one conversion script.


robjmcgibbon:
The conversion script fails because we need the upper limit functionality from velociraptor-python

PR #200, opened 2024-05-25
Update Driver data point

robjmcgibbon:
The previous value used here was from an unpublished version of the paper. We're now using the published value. There's no pipeline-configs pull request to go alongside this one since this value is found automatically by https://github.com/SWIFTSIM/pipeline-configs/blob/686a00505da26360b3f7814fc7736ddc795972c9/colibre/scripts/stellar_mass_evolution.py#L68

stellar_mass_evolution


data/StellarMassDensity/conversion/convertDriver2022.py
```diff
@@ -13,25 +13,19 @@
...
-
-# Convert to densities assuming the simulation's cosmology
-rho_crit = unyt.unyt_array.from_astropy(cosmology.critical_density0).to("Msun/Mpc**3")
-stellar_mass_density = unyt.unyt_array([Omega_star * rho_crit])
-stellar_mass_density_p = unyt.unyt_array([Omega_star_p * rho_crit])
-stellar_mass_density_m = unyt.unyt_array([Omega_star_m * rho_crit])
-
-# Construct the error bar
-y_scatter = unyt.unyt_array((stellar_mass_density_m, stellar_mass_density_p))
+# Value taken from section 6.2 of the paper
```
EvgeniiChaikin, line 22: Maybe it is good to mention that the value used here includes

-- a correction accounting for the re-normalization of GAMA to SDSS (Section 6.1)
-- a correction for the evolution of stellar mass density to a z = 0.0 (Section 6.2)

EvgeniiChaikin (approved):
Thanks @robjmcgibbon !

All looks good! I have added one small comment aiming to clarify the two corrections that are used by Driver+2022 to compute `rho_star`

data/StellarMassDensity/conversion/convertDriver2022.py
robjmcgibbon: Added!

PR #201, opened 2024-08-13
Add z-dependent dust mass function

robjmcgibbon:
Adds the dust mass function from https://ui.adsabs.harvard.edu/abs/2024arXiv240709607T/abstract

z2
z6


robjmcgibbon:
I've also added some data from https://arxiv.org/abs/2408.03996v1. See https://leiden-observatory.slack.com/archives/G035RUR984V/p1723648581438459

